PR #19, opened 2021-03-25
Attempt at finding discrepancy between main and adapt-for-rebar3 (no -otp-)

paulo-ferraz-oliveira:
⚠️ this is not a mergeable. This is to fundament an ongoing topic discussion.

@ericmj: I created this very small PoC to see try and find the discrepancy from `main` and #9.

I target (as per our discussion) `ubuntu-16.04`, Elixir 1.2.6 and Erlang/OTP 19.3.6.

I also try to output the actual installed versions of Erlang and Elixir.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,42 +1,18 @@
 name: Test
 
 on:
-  push: {branches: main}
-  pull_request: {branches: main}
-  repository_dispatch:
-  workflow_dispatch:
+  pull_request: {branches: '*'}
 
 jobs:
-  unit_test:
-    runs-on: ubuntu-latest
-    name: Unit tests
-    steps:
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
Fails at https://github.com/erlef/setup-elixir/runs/2195970211?check_suite_focus=true#step:3:626, even though the error is hidden (probably the script exits with 0).

This is something that I'd fixed in my pull request (the fact that no such error goes unnoticed) which is why it's revealed as such.

paulo-ferraz-oliveira:
I'm starting to think some of these issues are env. path -related.

paulo-ferraz-oliveira:
Discrepancy found. It was bad path manipulation. Closing.

PR #24, opened 2021-03-30
Apply stricter linting

paulo-ferraz-oliveira:
So, after the previous experience with #9 I thought we could further restrict the JavaScript code by making for stricter linting. This pull request presents that.

How it rolled:

1. I moved from `google` to `airbnb` (https://github.com/erlef/setup-beam/actions/runs/702116828)
2. I ran the CI command and started _fixing_ stuff

(this actually came from a comment by Eric where he asked if `return await` was OK - I thought "if it isn't there's probably some linting, out there, that'll help us further")

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -1,9 +1,9 @@
 const core = require('@actions/core')
-const {exec} = require('@actions/exec')
-const installer = require('./installer')
+const { exec } = require('@actions/exec')
 const path = require('path')
 const semver = require('semver')
 const https = require('https')
+const installer = require('./installer')
 
 main().catch((err) => {
   core.setFailed(err.message)
@@ -13,10 +13,10 @@ async function main() {
   installer.checkPlatform()
 
   const osVersion = getRunnerOSVersion()
-  const otpSpec = core.getInput('otp-version', {required: true})
+  const otpSpec = core.getInput('otp-version', { required: true })
   const otpVersion = await installOTP(otpSpec, osVersion)
 
-  const elixirSpec = core.getInput('elixir-version', {required: false})
+  const elixirSpec = core.getInput('elixir-version', { required: false })
   const elixirInstalled = await maybeInstallElixir(elixirSpec, otpVersion)
   if (elixirInstalled === true) {
     const shouldMixRebar = core.getInput('install-rebar', {
@@ -29,7 +29,7 @@ async function main() {
     await mix(shouldMixHex, 'hex')
   }
 
-  const rebar3Spec = core.getInput('rebar3-version', {required: false})
+  const rebar3Spec = core.getInput('rebar3-version', { required: false })
   maybeInstallRebar3(rebar3Spec)
 }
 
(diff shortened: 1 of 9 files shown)


src/setup-beam.js
```diff
@@ -133,15 +133,15 @@ async function getElixirVersion(exSpec0, otpVersion) {
       `Requested Elixir version (from spec ${exSpec0}) not found in build listing`,
     )
   }
-  const otpMatch = otpVersion.match(/^(?:OTP-)?([^\.]+)/)
+  const otpMatch = otpVersion.match(/^(?:OTP-)?([^.]+)/)
```
paulo-ferraz-oliveira, line 136: Not a real problem, but not visible with the previously linting procedure, either.

src/setup-beam.js
```diff
@@ -179,8 +179,7 @@ async function getOTPVersions(osVersion) {
   otpVersionsListing
     .trim()
     .split('\n')
-    .map((line) => {
-      debugger
```
paulo-ferraz-oliveira, line 183: Not a real problem, but not visible with the previous linting procedure, either.

ericmj (approved):

src/setup-beam.js
```diff
@@ -179,14 +179,12 @@ async function getOTPVersions(osVersion) {
...
+    .forEach((line) => {
```
paulo-ferraz-oliveira, line 182: `.map` not really required

src/setup-beam.js
```diff
@@ -179,14 +179,12 @@ async function getOTPVersions(osVersion) {
...
       const otpMatch = line.match(/^(OTP-)?([^ ]+)/)
 
       const otpVersion = otpMatch[2]
       otpVersions.set(otpVersion, otpMatch[0]) // we keep the original for later reference
     })
-    .sort()
```
paulo-ferraz-oliveira, line 189: `.sort` not really doing anything (apart from crashing if we use `.forEach`)

starbelly (approved):

paulo-ferraz-oliveira:
I'm still getting the following messages:

"
Check licenses Expected — Waiting for status to be reported
Required
OTP 21.x / Elixir <1.9.1 Expected — Waiting for status to be reported
Required
OTP 22.0 / Elixir 1.9.1 Expected — Waiting for status to be reported
Required
OTP 22.0 / Elixir master Expected — Waiting for status to be reported
Required
OTP 23.0 / Elixir 1.11.0-rc.0 Expected — Waiting for status to be reported
Required
OTP master / Elixir 1.11.3 Expected — Waiting for status to be reported
Required
Unit tests Expected — Waiting for status to be reported
Required
"

in the above box

paulo-ferraz-oliveira:
I'm not sure this is what's preventing me from merging.

starbelly:
@paulo-ferraz-oliveira yes, this would prevent merging. I'd rather not disable it, but if this is going to be a consistent problem 🤔 

The question is why? Why is it not running?

starbelly:
I see the issue...

starbelly:
@paulo-ferraz-oliveira so the issue was these outdated / no longer existent checks were still marked as required per our branch rules. Believe everything is up to date now. If you do a force push we can make sure that is so.

starbelly:
What's annoying is github adds every single test in the pre-release integration matrix as an individual status check. I have temporarily unchecked all those and will revisit after this is merged. Rationale is it's not easy to decipher some of the old status checks from the new.

PR #38, opened 2021-05-26
Test with no rebar3 installation

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
Tests done. It seems the Ubuntu image pull Erlang in with `rebar3`.

PR #49, opened 2021-06-22
Add Windows to the mix

### LINKED ISSUE #12: Have the action work on Windows builds (from setup-erlang)
At the moment it's Linux only.

### LINKED ISSUE #44: With .../ubuntu-20.04/builds.txt 503'ing error message is out of scope
The error you get is not from the fact the file is not available, which it should be, but further down the execution, by stating the version is not found in the listing.

When trying to download whatever file or fetching whatever external dependency we shouldn't hide errors behind default return values, meaning we should crash whenever https://github.com/erlef/setup-beam/blob/main/src/setup-beam.js#L272 is invoked.

paulo-ferraz-oliveira:
Opening as draft.

As-is, stuff appears to be working (and you're Ok to review, since I don't expect this to change much) but a couple of questions reside. I'll comment on the code later.

Closes #12.
Closes #44.

**Edit**: one big question for me is "Should we open an exception to the Windows installation script in order to support an even further past"?


.eslintrc.yml
```diff
@@ -6,14 +6,17 @@ extends:
...
+  no-unused-vars: ["error", {"varsIgnorePattern": "^_"}]
+  no-use-before-define: 0
+  operator-linebreak: 0
+  require-jsdoc: 0
   semi: 0
-}
+
+settings:
+  react:
+    version: 999.999.999
```
paulo-ferraz-oliveira, line 22: Prevent a warning that makes no sense

.github/workflows/ci.yml
```diff
@@ -21,12 +21,12 @@ jobs:
       - run: npm install -g npm
       - run: npm install
       - run: npm run build
+      - run: npm run format
```
paulo-ferraz-oliveira, line 24: Changing this always seems like a game of hide-and-seek.

.github/workflows/test.yml
```diff
@@ -2,10 +2,12 @@
 name: test
 
 on:
-  push: {branches: main}
-  pull_request: {branches: main}
-  repository_dispatch:
-  workflow_dispatch:
+  push:
```
paulo-ferraz-oliveira, line 5: Made it the same as `ci.yml`

.github/workflows/test.yml
```diff
@@ -136,3 +138,43 @@ jobs:
...
+      Pre-release integration tests
+      (Windows ${{matrix.combo.os}},
+       Erlang/OTP ${{matrix.combo.otp-version}},
+       rebar3 ${{matrix.combo.rebar3-version}})
+    runs-on: ${{matrix.combo.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        combo:
+          - otp-version: '24.0.2'
```
paulo-ferraz-oliveira, line 153: Let me know if these are enough

.github/workflows/test.yml
```diff
@@ -136,3 +138,43 @@ jobs:
...
+          otp-version: ${{matrix.combo.otp-version}}
+          rebar3-version: ${{matrix.combo.rebar3-version}}
+      - name: Erlang/OTP version (action)
+        run: echo "Erlang/OTP ${{steps.setup-beam.outputs.otp-version}}"
+      - name: rebar3 version (action)
+        run: echo "rebar3 ${{steps.setup-beam.outputs.rebar3-version}}"
+      - name: Run rebar3 project tests
+        run: |
+          cd test-projects/rebar3
+          & rebar3 ct
```
paulo-ferraz-oliveira, line 180: Let me know if this interface is acceptable (rebar3 here is actually a `.ps1` file, in the path)

README.md
```diff
@@ -45,6 +45,8 @@ and Erlang/OTP.
 | ubuntu-16.04 | 17 - 24    | ✅
 | ubuntu-18.04 | 17 - 24    | ✅
 | ubuntu-20.04 | 20 - 24    | ✅
+| windows-2016 | 23 - 24    | ✅
```
paulo-ferraz-oliveira, line 48: Yeah 😢 
I mean, we could try and fetch 22, for example, but then it's not saved in the same "C:\Program Files\" folder, so I prefer to do this initially, to have no exception, and ask for your opinion on this choice

__tests__/setup-beam.test.js
```diff
@@ -81,47 +81,69 @@ async function testOTPVersions() {
   let spec
   let osVersion
 
-  spec = '19.3.x'
```
paulo-ferraz-oliveira, line 84: This just got encapsulated inside an `if`.

dist/install-elixir.ps1
```diff
@@ -0,0 +1,5 @@
+Write-Output "Installer for Elixir for Windows not available"
```
paulo-ferraz-oliveira, line 1: He/she/it who comes after is more than welcome to take a stab at this. For the time being I preferred to be explicit in saying Elixir is not available.

src/install-elixir.ps1
```diff
@@ -0,0 +1,5 @@
+Write-Output "Installer for Elixir for Windows not available"
```
paulo-ferraz-oliveira, line 1: He/she/it who comes after is more than welcome to take a stab at this. For the time being I preferred to be explicit in saying Elixir is not available.

src/install-otp.ps1
```diff
@@ -0,0 +1,19 @@
+param([Parameter(Mandatory=$true)][string]$VSN)
```
paulo-ferraz-oliveira, line 1: All of this is basically a translation of `install-otp.sh`. I'm sure there's a better way to write a PowerShell script, but my knowledge here is very limited. I just wanted to make this work.

src/install-rebar3.ps1
```diff
@@ -0,0 +1,24 @@
...
+$DIR_FOR_BIN=".setup-beam/rebar3"
+
+Remove-Item -Force "$FILE_OUTPUT" -ErrorAction SilentlyContinue
+Remove-Item -Recurse -Force "$DIR_FOR_BIN" -ErrorAction SilentlyContinue
+$ProgressPreference="SilentlyContinue"
+Invoke-WebRequest "https://github.com/erlang/rebar3/releases/download/${VSN}/${FILE_INPUT}" -OutFile "$FILE_OUTPUT"
+$ProgressPreference="Continue"
+New-Item "$DIR_FOR_BIN/bin" -ItemType Directory | Out-Null
+Move-Item "$FILE_OUTPUT" "$DIR_FOR_BIN/bin"
+Write-Output "& escript.exe $PWD/$DIR_FOR_BIN/bin/$FILE_OUTPUT `$args" | Out-File -FilePath "$FILE_OUTPUT_PS1" -Encoding utf8 -Append
```
paulo-ferraz-oliveira, line 19: This is what actually creates the `.ps1` that'll be included in the path.

src/setup-beam.js
```diff
@@ -17,20 +17,24 @@ async function main() {
   const otpVersion = await installOTP(otpSpec, osVersion)
 
   const elixirSpec = core.getInput('elixir-version', { required: false })
-  const elixirInstalled = await maybeInstallElixir(elixirSpec, otpVersion)
+  const shouldMixHex = core.getInput('install-hex', {
```
paulo-ferraz-oliveira, line 20: Moved here, since I was getting `install-hex: true` in the logs while it wasn't happening, so I wanted to make the default override explicit.

src/setup-beam.js
```diff
@@ -17,20 +17,24 @@ async function main() {
...
     await mix(shouldMixRebar, 'rebar')
-    const shouldMixHex = core.getInput('install-hex', {
-      required: false,
-    })
     await mix(shouldMixHex, 'hex')
   }
 
   const rebar3Spec = core.getInput('rebar3-version', { required: false })
-  maybeInstallRebar3(rebar3Spec)
+  await maybeInstallRebar3(rebar3Spec)
```
paulo-ferraz-oliveira, line 37: This was provoking `UnhandledPromiseRejectionWarning`.

src/setup-beam.js
```diff
@@ -175,34 +189,61 @@ async function getRebar3Version(r3Spec) {
...
-  const otpVersions = new Map()
+  let originListing
+  let pageIdxs
+  if (process.platform === 'linux') {
+    originListing = `https://repo.hex.pm/builds/otp/${osVersion}/builds.txt`
+    pageIdxs = [null]
+  } else if (process.platform === 'win32') {
+    originListing =
+      'https://api.github.com/repos/erlang/otp/releases?per_page=100'
+    pageIdxs = [1, 2, 3]
```
paulo-ferraz-oliveira, line 200: Pagination, even if very explicit... I didn't want to deal with looping until I got 404 or `[]`. This seems enough, for the time being.

src/setup-beam.js
```diff
@@ -175,34 +189,61 @@ async function getRebar3Version(r3Spec) {
...
-      otpVersions.set(otpVersion, otpMatch[0]) // we keep the original for later reference
+  if (process.platform === 'linux') {
+    otpVersionsListings
+      .trim()
+      .split('\n')
+      .forEach((line) => {
+        const otpMatch = line.match(/^(OTP-)?([^ ]+)/)
+
+        let otpVersion = otpMatch[2]
+        if (semver.validRange(otpVersion) && hasPatch(otpVersion)) {
```
paulo-ferraz-oliveira, line 214: I added `hasPatch` so that eg. `23.2` doesn't get transformed into `23.2.0`, which doesn't exist.

src/setup-beam.js
```diff
@@ -175,34 +189,61 @@ async function getRebar3Version(r3Spec) {
...
+          otpVersion = semver.minVersion(otpVersion).version
+        }
+        otpVersions.set(otpVersion, otpMatch[0]) // we keep the original for later reference
+      })
+  } else if (process.platform === 'win32') {
+    otpVersionsListings.forEach((otpVersionsListing) => {
+      JSON.parse(otpVersionsListing)
+        .map((x) => x.assets)
+        .flat()
+        .filter((x) => x.name.match(/^otp_win64_.*.exe$/))
```
paulo-ferraz-oliveira, line 224: `win64` here, `win32` above. Consistency for the win 🤕

src/setup-beam.js
```diff
@@ -241,43 +286,79 @@ function getVersionFromSpec(spec, versions) {
...
     ubuntu20: 'ubuntu-20.04',
+    win16: 'windows-2016',
+    win19: 'windows-2019',
   }
 
-  return mapToUbuntuVersion[process.env.ImageOS] || 'ubuntu-18.04'
+  return ImageOSToContainer[process.env.ImageOS]
 }
 
-function get(url) {
```
paulo-ferraz-oliveira, line 253: Adapted in order to support pagination.

paulo-ferraz-oliveira:
Comments done. CI passing. Moving to Ready for review.

paulo-ferraz-oliveira:
@wojtekmach, I'd like to invite you as a reviewer, but the UI won't let me.

README.md
```diff
@@ -130,6 +132,26 @@ jobs:
...
+### Basic example (`rebar3` on Windows 2016)
+
+```yaml
+# create this in .github/workflows/ci.yml
+on: push
+
+jobs:
+  test:
+    runs-on: windows-2019
+    name: Erlang/OTP ${{matrix.otp}} / rebar3 ${{matrix.rebar3}}
```
wojtekmach, line 144: this workflow doesn't specify a matrix, so `matrix.otp` etc is invalid here. I'd just leave this out from the snippet.

```suggestion
```

README.md
```diff
@@ -130,6 +132,26 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: '24.0.2'
+          rebar3-version: '3.16.1'
+      - shell: pwsh
+        run: "& rebar3 ct"
```
wojtekmach, line 152: this can also be:

```suggestion
      - run: rebar3 ct
```

(see: https://github.com/wojtekmach/test_erlang/blob/1fe6ff4c8bd491e7911e4207c8a73acb58c04fde/.github/workflows/main.yml#L32 & https://github.com/wojtekmach/test_erlang/runs/2888692249?check_suite_focus=true)

wojtekmach (approved):
This is fantastic <3 I was able to test it on a an example project and it works flawlessly. 

I'm not well versed in JS or windows scripting, so I focused on the outside behaviour and not the implementation.

README.md
wojtekmach: 23+ sounds fine to me. I'd even make it 24+ if that'd simplify anything. As a matter of fact, I'd make it windows-2019 & 24+ if that'd simplify anything. We literally didn't have any other options for Windows so this more than what we could hope for anyway.

dist/index.js
```diff
@@ -4698,6 +4726,12 @@ async function maybeInstallElixir(elixirSpec, otpVersion) {
     return true
   }
 
+  if (shouldMixHex) {
+    console.log(
+      "hex will not be installed (overriding default) since Elixir wasn't either",
```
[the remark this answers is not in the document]
wojtekmach: ah, I guess this is related to https://github.com/erlef/setup-beam/pull/49#discussion_r656447109?

so someone set `install-hex: true` but left `elixir-version` out? to me that should be an error: "cannot set install-hex: true without specifying elixir-version" or something.

.github/workflows/test.yml
starbelly: Quite acceptable! It will be very much appreciated by users.

src/install-elixir.ps1
starbelly: It's also not clear whether the elixir web installer can do unattended installs to boot.

starbelly:
@paulo-ferraz-oliveira Did an initial first pass, looks solid! Amazing stuff 🚀 Will review in anger later this evening.

src/install-elixir.ps1
wojtekmach: I thought the main selling point of Elixir web installer was that it would install OTP for you as well. Which is not what we want as we want more control into particular OTP versions.

I think we could treat Elixir tarball similar to Erlang tarballs, we unpack it somewhere and add bin/ to path. Elixir ships with Windows scripts (https://github.com/elixir-lang/elixir/tree/master/bin). I don't know if it is relevant but it looks like it only has .ps1 for mix and not others.

THat being said, probably out of the scope of this PR.

starbelly: Ahh yeah, that makes sense. I've never installed elixir on windows, but yeah I would think the precompiled tarball should "just work" ™️

dist/index.js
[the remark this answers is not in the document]
starbelly: Agreed, makes no sense to have one without the other.

starbelly (approved):

jclem:
Just dropping by to say how happy I am to see PRs like this. I would not have had the time or probably the expertise personally to get this done (even with outside contributors) while I was maintaining this action under the official Actions org. Thanks again for agreeing to take this on, Erlef folks!

starbelly:
@paulo-ferraz-oliveira approved, but I don't understand this question : 

> Should we open an exception to the Windows installation script in order to support an even further past?

README.md
paulo-ferraz-oliveira: > I'd even make it 24+ if that'd simplify anything.

It wouldn't 😄. I started with 19, actually, and most of the process works (download, install, etc.), but then, since the path is not easy to figure out - I don't know the equivalent of `find` in PowerShell and was lazy to try it out - I started bumping it until I came to this.

src/install-elixir.ps1
paulo-ferraz-oliveira: > THat being said, probably out of the scope of this PR.

If it very simple to add (ie.I have instructions or pointers) I don't mind doing it - in that sense it's not out-of-scope because Elixir was part of the mix already, but I don't have an easy way to do it, since I might miss something. With Erlang/OTP I knew exactly what I was looking for.

paulo-ferraz-oliveira: That said, can any of you give me pointers as to how this could be achieved?

.github/workflows/test.yml
```diff
@@ -2,10 +2,12 @@
 name: test
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Yeah, they could be different jobs, sure. I never bothered to change this since they were already separated from the beginning. I guess it's for semantics purposes 🤷‍♂️.

As I see it (but this is me only, I guess) `ci` will do the "unit tests". `test` will do something else. (and before these changes one of the two could actually be launched by pressing a button on the UI - it's those `_dispatch` options).

I can join them. @starbelly, do you feel it's better if these go to the same file now that they're dispatched in the same conditions?

.github/workflows/test.yml
```diff
@@ -136,3 +138,43 @@ jobs:
           cd test-projects/rebar3
           rebar3 ct
         if: ${{matrix.combo.rebar3-version}}
+
+  integration_test_windows:
+    name: >
+      Pre-release integration tests
+      (Windows ${{matrix.combo.os}},
+       Erlang/OTP ${{matrix.combo.otp-version}},
+       rebar3 ${{matrix.combo.rebar3-version}})
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: This looks better. Let me just get @starbelly's input on the previous question then I'll handle them at the same time.

> I'm not sure why rebar3-version didn't show up in the screenshot above.

Me neither: checking... well, in that case we didn't choose a `rebar3` on purpose, and I couldn't figure out how to have conditional titles, so that's the best I could do. We could just remove that info and put it in the step name, for example (not tested yet).

dist/index.js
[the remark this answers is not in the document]
paulo-ferraz-oliveira: > so someone set install-hex: true but left elixir-version out?

The fact is it was a "no-choice". `install-hex` is `true` by default (I think it's been like this for a while - didn't check). I can:
1. improve the warning/error messages, or
2. make it default in the code, not as an input.

Whatever choice I agree this could do with some tweaking for clarity. Let me open a new issue.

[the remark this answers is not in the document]
paulo-ferraz-oliveira: https://github.com/erlef/setup-beam/issues/50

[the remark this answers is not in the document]
wojtekmach: ohhhh, install-hex is true by default, that makes sense. in that case i would honestly silently ignore this and call it a day.

README.md
paulo-ferraz-oliveira: Oh, this is so much better. Let me do that.

README.md
paulo-ferraz-oliveira: Commited. Thanks.

README.md
paulo-ferraz-oliveira: I'll also try this approach in `test.yml`. I didn't know `shell` was a thing. Have not completely read the specification, so am working on a per-need basis 😄

paulo-ferraz-oliveira:
@starbelly 

> Should we open an exception to the Windows installation script in order to support an even further past?

We're only going down to 23, but there's installers for previous versions. The thing is those don't follow the same pattern for the "C:\Program Files\" path (which, mind you, could also break in the future, since it's not controlled by us). As long as I could be sure that `erl.exe` was there and I could do a *nix-style `find` we could support more versions (as per the compatibility list - I'm sure people'll always try to use in what they need first). I'll keep this in the backlog and implement if I can find an easy solution. (it was too late in the evening when I did the first implementation - which had no `rebar3` yet, and I was too tired to look for anything else)

.github/workflows/test.yml
paulo-ferraz-oliveira: And with @wojtekmach's help, even further simplified: 23d923c.

paulo-ferraz-oliveira:
I'll wait for your comments on the Unresolved conversations, but have started a new issue for a follow-up PR, lest we cram everything here in a single commit 🤣

README.md
wojtekmach: cool, lets leave things as is.

src/install-elixir.ps1
wojtekmach: This seems to do the trick!

https://github.com/wojtekmach/test_erlang/blob/aa81bbbc7654af98adba2583061179ffd1b60624/.github/workflows/main.yml#L25:L40

https://github.com/wojtekmach/test_erlang/runs/2894176893?check_suite_focus=true

wojtekmach: we should of course download from repo.hex.pm, same as we do for Linux, because we want to grab Elixir compiled against a matching OTP version.

in fact, do we even need `install-elixir.ps1`? perhaps it's easier to do everything in JS because it is supposed to work cross-platform, all we need to do is unzip and add the bin/ directory to $GITHUB_PATH. Again I'm not well versed to comment on the implementation, pick whatever is the simplest and most maintainable.

paulo-ferraz-oliveira: Yeah, the other day I was wondering why some stuff was written as Shell script vs JavaScript, myself. I think it should only be so when execution is needed in the target platform (e.g. I want to get the Elixir version from Elixir to make sure the installation process was Ok) - but most other stuff could be converted to JS... anyway, I've linked this to the newer issue, so your comments will be _ported_ over also. Thanks for the help.

paulo-ferraz-oliveira:
I'm merging this to start working on #51. This is usable as-is, in any case. Feel free to comment further here or in #51 and I'll do my best to approach your suggestions.

PR #59, opened 2021-07-09
Improve warning message for missing ImageOS env. variable (self-hosted runners)

### LINKED ISSUE #58: Action incompatible with self-hosted runners
Hi,

This action is now incompatible with self-hosted runners, where the `ImageOS` environment variable is not provided. To let it work, the environment variable has to be injected manually:

```yaml
      - uses: erlef/setup-beam@v1
        with:
          otp-version: ${{ env.TARGET_ERLANG_VERSION }}
          elixir-version: ${{ env.TARGET_ELIXIR_VERSION }}
        env:
          # See:
          # https://github.com/erlef/setup-beam/blob/master/src/setup-beam.js#L340
          ImageOS: ubuntu20
```

Honestly, I do not know what to think about this 😞 , maybe you should rework this part by asking the system directly the version with `uname` maybe :).

paulo-ferraz-oliveira:
Closes #58.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: '12'
+          node-version: 12
       - run: npm install -g npm
       - run: npm install
       - run: npm run build
@@ -31,12 +31,24 @@ jobs:
       - name: Check if build left artifacts
         run: git diff --exit-code
 
-  unit_test:
-    name: Unit tests
+  unit_tests_ubuntu:
+    name: Unit tests (Ubuntu)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
-        with: {node-version: '12'}
+        with:
+          node-version: 12
       - run: npm ci
+      - run: npm test
+
+  unit_tests_windows:
+    name: Unit tests (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install --production
       - run: npm test
--- a/README.md
+++ b/README.md
@@ -53,6 +53,35 @@ and Erlang/OTP.
 
 **Note** *: prior to 23, Windows builds are only available for minor versions, e.g. 21.0, 21.3, 22.0, etc.
 
+### Self-hosted runners
+
+Self-hosted runners need to set env. variable `ImageOS` to one of the following, since the action
+uses that to download assets:
+
+| ImageOS  | Operating system
+|-         |-
+| ubuntu16 | ubuntu-16.04
+| ubuntu18 | ubuntu-18.04
+| ubuntu20 | ubuntu-20.04
+| win16    | windows-2016
+| win19    | windows-2019
+
+as per the following example:
+
+```yaml
+...
+
+jobs:
+  test:
+    runs-on: self-hosted
+    env:
+      ImageOS: ubuntu20 # equivalent to runs-on ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        ...
+```
+
 ### Basic example (Erlang/OTP + Elixir, on Ubuntu)
 
 ```yaml
--- a/__tests__/setup-beam.test.js
+++ b/__tests__/setup-beam.test.js
@@ -136,7 +136,7 @@ async function testOTPVersions() {
 
     spec = '23.2.x'
     osVersion = 'windows-2016'
-    expected = '23.2.7.4'
+    expected = '23.2.7'
     got = await setupBeam.getOTPVersion(spec, osVersion)
     assert.deepStrictEqual(got, expected)
 
--- a/dist/index.js
+++ b/dist/index.js
@@ -5001,8 +5001,19 @@ function getRunnerOSVersion() {
     win16: 'windows-2016',
     win19: 'windows-2019',
   }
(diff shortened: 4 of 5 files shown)


.github/workflows/action.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: '12'
```
paulo-ferraz-oliveira, line 20: Not strictly required for this pull request

.github/workflows/action.yml
```diff
@@ -31,12 +31,24 @@ jobs:
...
 
-  unit_test:
-    name: Unit tests
+  unit_tests_ubuntu:
+    name: Unit tests (Ubuntu)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
-        with: {node-version: '12'}
```
paulo-ferraz-oliveira, line 40: Not strictly required for this pull request

.github/workflows/action.yml
```diff
@@ -31,12 +31,24 @@ jobs:
...
+        with:
+          node-version: 12
       - run: npm ci
       - run: npm test
+
+  unit_tests_windows:
```
paulo-ferraz-oliveira, line 45: We were actually not testing this on Windows, and this change helped find a bug in the tests 😄

.github/workflows/action.yml
```diff
@@ -31,12 +31,24 @@ jobs:
...
+    name: Unit tests (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install --production
```
paulo-ferraz-oliveira, line 53: If I do `npm ci` `devDependencies` try to get installed, and this fails because (at least) of `shellcheck`

README.md
```diff
@@ -53,6 +53,36 @@ and Erlang/OTP.
 
 **Note** *: prior to 23, Windows builds are only available for minor versions, e.g. 21.0, 21.3, 22.0, etc.
 
+### Self-hosted runners
+
+We try to guess the self-hosted runner _container_ equivalent (to download assets used in the action)
+by making direct calls to the operating system, but this might fail, in which case we
+warn with an error message, and you should set environment variable `ImageOS` to one of the following:
```
paulo-ferraz-oliveira, line 60: Still need to run a short test using this action to make sure the error message is not an "unhandled promise".

README.md
```diff
@@ -53,6 +53,36 @@ and Erlang/OTP.
...
+as per the following example:
+
+```yaml
+...
+
+jobs:
+  test:
+    runs-on: self-hosted
+    env:
+      ImageOS: ubuntu20 # equivalent to runs-on ubuntu-20.04
```
paulo-ferraz-oliveira, line 79: This is our safety net in case the "convoluted" guess doesn't work as expected (it's tested in all versions the action supports, as of now)

__tests__/setup-beam.test.js
```diff
@@ -136,7 +137,7 @@ async function testOTPVersions() {
 
     spec = '23.2.x'
     osVersion = 'windows-2016'
-    expected = '23.2.7.4'
```
paulo-ferraz-oliveira, line 139: Testing it on Windows surfaced this

src/setup-beam.js
```diff
@@ -338,15 +339,51 @@ function sortVersions(left, right) {
...
+  return {
     ubuntu16: 'ubuntu-16.04',
     ubuntu18: 'ubuntu-18.04',
     ubuntu20: 'ubuntu-20.04',
     win16: 'windows-2016',
     win19: 'windows-2019',
+  }[process.env.ImageOS]
+}
+
+function getContainerFromOS() {
```
paulo-ferraz-oliveira, line 355: This is where the guessing game now happens. It's convoluted, sure, but I'm not sure there's a better way to fetch this info. from the target OSs

README.md
paulo-ferraz-oliveira: Test is done. Error is caught as expected.

src/setup-beam.js
starbelly: See my comment here https://github.com/erlef/setup-beam/issues/58#issuecomment-877672764

paulo-ferraz-oliveira: Per "I'd rather us not start doing guess work around operating systems" I'm reverting this guessing bit. Whatever it is, the self-hosted runner will have to use the `ImageOS` env. variable to achieve the same behaviour. I surely don't want to introduce a new env. variable and that one's already documented in the README.

README.md
paulo-ferraz-oliveira: This comment no longer applies, since there's no guesswork

dist/index.js
```diff
@@ -5001,8 +5001,19 @@ function getRunnerOSVersion() {
     win16: 'windows-2016',
     win19: 'windows-2019',
   }
+  const containerFromEnvImageOS = ImageOSToContainer[process.env.ImageOS]
 
-  return ImageOSToContainer[process.env.ImageOS]
+  if (!containerFromEnvImageOS) {
+    throw new Error(
+      "Tried to map a target OS from env. variable 'ImageOS', but failed. If you're using a " +
```
paulo-ferraz-oliveira, line 5008: Example output: 
```
Tried to map a target OS from env. variable 'ImageOS', but failed. If you're using a self-hosted runner, you should set 'env': 'ImageOS': ... to one of the following: ['ubuntu16', 'ubuntu18', 'ubuntu20', 'win16', 'win19']
```

paulo-ferraz-oliveira: (this way the list will be as updated as the constant definition)

starbelly (approved):

PR #63, opened 2021-07-25
Ease debugging around ImageOS env.var

paulo-ferraz-oliveira:
We output the variable's value to ease debugging around this issue.

Ref.: https://github.com/erlef/setup-beam/issues/58#issuecomment-886125419


ericmj (approved):

paulo-ferraz-oliveira:
Cool. Thanks.

PR #64, opened 2021-07-26
Add tests for github actions problem matchers

btkostner:
I figured I would setup testing for the problem matchers before I start adding some. As requested in https://github.com/erlef/setup-beam/issues/62#issuecomment-886267969

PR #65, opened 2021-07-27
Add credo and dialyzer problem matchers

btkostner:
Waiting on #64 so I can add tests for these two. Both of these are setup to parse the _default_ output of both CLI commands in elixir. You can include `mix credo` and `mix dialyzer` in a GitHub action and the output should be parsed correctly. You can see the results on a test repo I was using:

Credo: https://github.com/btkostner/setup-beam-action-test/commit/8af098c62db381e123e72928886d5449d470a101
Dialyzer: https://github.com/btkostner/setup-beam-action-test/commit/a9bc2839e5c4b68d8191bcb36c1b59c0d29ec871

I'm pretty happy with the credo regex specificity, the dialyzer one seems a little generic, but I haven't had any false positives yet. Let me know if there are improvements you can think of.

Also closes: https://github.com/rrrene/credo/pull/870


paulo-ferraz-oliveira:
I'm pulling in a few other reviewers that'll be more experienced in Elixir than me. In any case, this won't be too Elixir-specific, I think.

paulo-ferraz-oliveira:
I think rebasing on top of `main` should make those artifact-related errors go away.

paulo-ferraz-oliveira:
This looks Ok to me. Any chance you could run it on top of the test repo. you'd created before and post a link to the results here?

btkostner:
Seems there was an issue with the Credo regex parsing the severity. I updated it and have opened up a PR on my test repo to show the results:

https://github.com/btkostner/setup-beam-action-test/pull/1/files

paulo-ferraz-oliveira:
Does that cover `dialyzer` also, your repo.?

btkostner:
@paulo-ferraz-oliveira Yes. The first message is from Dialyzer. Then there is another one below that for
> The pattern can never match the type.

paulo-ferraz-oliveira:
Oh, right, it's written in the title 😄 Do you want to add more stuff, still?

btkostner:
Nope. I'm happy with it as is!

starbelly:
> I'm pulling in a few other reviewers that'll be more experienced in Elixir than me. In any case, this won't be too Elixir-specific, I think.

Correct, definitely not elixir specific at all. It seems to work and I don't see anything egregious. Bad case scenario would be the matchers get triggered for something they're not intended for and the regex should be adjusted, but that's non-harmful IMO. Worst case scenario if the first bad case scenario ensues, we observe we can't make regex that catch things appropriately and we have to remove problem matchers.

starbelly (approved):

paulo-ferraz-oliveira:
Squash-merging. Thanks for the pull request.

Issue #76, opened 2021-12-19
Build is failing per third-party license differences 


starbelly:
See https://github.com/erlef/setup-beam/runs/4572138656?check_suite_focus=true as an example.

paulo-ferraz-oliveira:
Hey.

You have to check the actions (I think it's `npm run licenses`) to update this locally, after which you have to upload the differences too.

We should probably have a "How to contribute" part in the README because some details (like this) aren't immediately obvious.

starbelly:
Yeah, I knew that part but I wasn't sure why there was a diff all of a sudden, I guess it just grabbed some latest versions in ci.

paulo-ferraz-oliveira:
That's the main part of it I don't like (this whole licensing thing - we inherited it, we didn't come up with it). This only proves that even for a given fixed version there are some internal moving parts (mostly some JS lib.).

Should we try to automate the update of that file? I'm sure whoever's doing it is not validating ALL the updates licenses one by one.

starbelly:
Well, I guess the question I ask myself is what do we get out of checking the licenses? I think perhaps made since when github owned the action as setup-elixir, but I don't think it makes sense for us.

paulo-ferraz-oliveira:
We check the licenses to make sure they're updated (as per the version we're pulling). Otherwise you might be importing a version of a dep. that has a non-proper (updated) license. This only happens when the license changes (or is added or removed), I imagine. But I'm Ok with whatever is decided for this. No real opinion on having the licenses declared or not (don't know if it's a legal requirement).

paulo-ferraz-oliveira:
@starbelly, do you think it's relevant to keep the generated 3rd party licenses next to the software? I mean, I don't see this being done often, but there might be good reason for this to have happened from the get-go.

If not, we should get rid of everything 3rd party license -related and avoid this class of issues.

starbelly:
Actually yeah, per the licenses of some of the third party deps, they must be included since pulling down the action from github ci is considered distribution.

paulo-ferraz-oliveira:
Ok, do you know how to automate this? 😄 I mean, we can always do it just before a release, but if automatic, it would probably be better.

starbelly:
@paulo-ferraz-oliveira Yeah, I think it's best if done before a release to ease the friction. We could setup scheduled builds to keep us in check.

paulo-ferraz-oliveira:
> We could setup scheduled builds to keep us in check.

[...]

PR #81, opened 2022-01-10
Make OTP installation optional

### LINKED ISSUE #79: Make OTP installation optional
Currently the `otp-version` argument is required. In some scenarios (e.g. when using custom OTP build from a different source), the action could still be useful - e.g. to install Elixir.

Would it make sense to make the `otp-version` argument optional, similarly to other versions?

michallepicki:
closes #79


paulo-ferraz-oliveira:
@michallepicki, thanks for this. Let me check the changes and get back to you.

.github/workflows/ubuntu.yml
```diff
@@ -155,3 +162,13 @@ jobs:
           cd test-projects/gleam_rebar3
           rebar3 eunit
         if: ${{matrix.combo.gleam-version}}
+      - name: Type check Gleam project
```
paulo-ferraz-oliveira, line 165:
```suggestion
      - name: Type check Gleam project (without installing Erlang/OTP)
```

.github/workflows/ubuntu.yml
```diff
@@ -155,3 +162,13 @@ jobs:
...
+        run: |
+          cd test-projects/gleam_gleam
+          gleam check
+        if: ${{ matrix.combo.gleam-version && !matrix.combo.otp-version }}
+      - name: Run Gleam project tests
```
paulo-ferraz-oliveira, line 170:
```suggestion
      - name: Run Gleam project tests (without `rebar3`)
```

src/setup-beam.js
```diff
@@ -37,7 +42,7 @@ async function main() {
   await maybeInstallRebar3(rebar3Spec)
 }
 
-async function installOTP(otpSpec, osVersion) {
+async function maybeInstallOTP(otpSpec, osVersion) {
```
paulo-ferraz-oliveira, line 45: Function name should not be `maybe...`. Execution is already made conditional by external factors.

src/setup-beam.js
```diff
@@ -15,19 +15,24 @@ async function main() {
...
+      const shouldMixRebar = core.getInput('install-rebar', {
+        required: false,
+      })
+      await mix(shouldMixRebar, 'rebar')
+      const shouldMixHex = core.getInput('install-hex', {
+        required: false,
+      })
+      await mix(shouldMixHex, 'hex')
+    }
+  } else if (elixirSpec) {
```
paulo-ferraz-oliveira, line 34: This should probably be related to Gleam, not Elixir (eg. `!gleamSpec`, or something), because the condition is "fail when no OTP if not Gleam" (otherwise every time we had a new language (eg. LFE) we'd have to come back to this.

Error can be something like "otp-version=false is only available when targeting Gleam".

src/setup-beam.js
michallepicki: Ah yes I was going back and forth with where to put the conditional. Will update.

test-projects/gleam_gleam/.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: test
+
+on:
+  push:
+    branches:
+      - master
```
paulo-ferraz-oliveira, line 6: We don't have a `master` branch on this repo.

michallepicki: that's a file generated by default by `gleam new`. I can remove this to avoid confusion

paulo-ferraz-oliveira:
Only a couple minor changes and we're almost there.

src/setup-beam.js
michallepicki: I added the error here to point out to the user that if they had put in both `otp-version` as `false` and `elixir`, `elixir` was not installed (because determining the version currently depends on `otp-version` being provided). So this is only specific to `elixir` and not gleam/lfe/etc.

michallepicki:
Not sure what this failure is about? https://github.com/erlef/setup-beam/runs/4762594908?check_suite_focus=true

src/setup-beam.js
paulo-ferraz-oliveira: Sure, but we want to warn the other way around (if you chose `otp-version: false` you can only target Gleam). Otherwise, tomorrow we include LFE in the action (and then Caramel, and then...) and have to touch that `if` again. I prefer the error to change to something like I stated "otp-version=false is only available for Gleam".

paulo-ferraz-oliveira:
Don't worry if the build is only failing for `Check if build left artifacts`, since this was not introduced by you.

paulo-ferraz-oliveira:
> Not sure what this failure is about? https://github.com/erlef/setup-beam/runs/4762594908?check_suite_focus=true

It's because the 3rd party licenses we're using were updated remotely but not locally (then the action updates them, checks against what you pushed, and errors out).

src/setup-beam.js
michallepicki: So you're saying there's a higher chance that other languages will have a "sub-version" dependency on OTP similar to Elixir than having the compiler written e.g. in Rust or OCaml.

I am not sure that will be the case. Maybe I'll just remove that `else` and error completely?

(Currently Caramel and Sesterl both are written in OCaml so if I make the change, and someone adds them, and wants to be able to type check without installing OTP, they will need to be added to the new `if` you're proposing)

michallepicki: I made the change you're proposing anyway :)

paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
I approve the current set of changes. Thanks much for your contribution 👍. I'd like to wait for @starbelly to review this, too, and maybe handle #76 before releasing a new minor.

src/setup-beam.js
michallepicki: No worries, I personally am not even sure this is worth merging if it's a special feature for Gleam only. But maybe Elixir could have a default fallback to oldest "compiled with" `otp-version` if not provided, then it could be useful as well for people using custom BEAM builds as well.

starbelly:
@michallepicki You should be able to rebase and we'll get a full passing build.

starbelly (approved):
This looks good to me, and I mean that :) 

Approving with the caveat that the build is still failing.

starbelly (changes requested):
Almost forgot, we need documentation for this. It should read clearly this option is currently only applicable to gleam.

README.md
```diff
@@ -180,9 +199,9 @@ The Elixir Problem Matchers in this repository are adapted from [here](https://g
 - `@v1.8`: the latest in the `1.8.z` series (this tag is movable),
 - `@v1.8.0`: release `1.8.0` (this tag is not movable).
 
-To prevent issues in CI (unless you're OK with potential incompatibility between versions - we
-make real a effort to not introduce those without bumping major) we suggest you to use `@vx.y.z`,
-whenever possible.
+We make a real effort to not introduce incompatibilities without changing the major
+version number. To be extra safe against changes causing issues in CI you can specify
+an exact version with `@vx.y.z`.
```
michallepicki, line 204: I found this part a bit confusing so I took liberty to change it. I think this is more clear. Or maybe I misunderstood and specifying all three numbers is _always_ recommended?

michallepicki (commented):
@starbelly Readme and action.yml updated!

README.md
```diff
@@ -149,7 +149,7 @@ jobs:
       - run: rebar3 ct
 ```
 
-### Example (Gleam + `rebar3`, on Ubuntu)
```
paulo-ferraz-oliveira, line 152: Do we want to get rid of the Gleam + `rebar3` example? I don't mind that much, but I don't understand why.

README.md
```diff
@@ -11,11 +11,11 @@
 This action sets up an Erlang/OTP environment for use in a GitHub Actions
 workflow by:
 
-- installing Erlang/OTP
```
paulo-ferraz-oliveira, line 14: ❤️

README.md
```diff
@@ -163,11 +163,30 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: false
+          gleam-version: '0.19.0-rc3'
+      - run: gleam check
 ```
 
+**Note**: The `otp-version: false` config is only applicable when installing Gleam.
```
paulo-ferraz-oliveira, line 188:
```suggestion
**Note**: the `otp-version: false` input is only applicable when installing Gleam.
```

paulo-ferraz-oliveira: Regarding the small caps, I'll be fixing the rest of the README in a different pull request.

README.md
paulo-ferraz-oliveira, line 204:
```suggestion
version number. To be extra safe against changes causing issues in your CI you should specify
an exact version with `@vx.y.z`.
```

paulo-ferraz-oliveira: It's not _always_ recommended, but it's preferable if you want a more stable environment (and even so, the underlying Node.js packages change, so it's really not just up to us, unfortunately). Your wording is better than what was there before, in any case.

paulo-ferraz-oliveira:
I left 2 comments and 2 code change suggestions. I'd wondered about the document changes, but had thought making the option visible would only bring more confusion. As is (current pull request), it might not.

README.md
michallepicki: Gleam switched to using its own binary as a build tool as the default, this likely will be the best supported way of building Gleam projects. And I thought it'd be weird having three examples specific to Gleam

paulo-ferraz-oliveira: Sure thing. That makes sense.

michallepicki:
Rebased from main to resolve conflict

paulo-ferraz-oliveira:
@michallepicki, so it's easier for me to follow up on what was request/done, resolve conversations only after you push the code changes (or even don't resolve them and let the one who opened them do). 👍

paulo-ferraz-oliveira:
I'm re-running all actions, after which I think we're good to merge.

paulo-ferraz-oliveira:
I believe the incomplete action runs are due to us still using `ubuntu-16.04`, but that's been handled in a different pull request, soon to be merged to `main`. I'm squash merging this, and once we have all the open pull requests merged and no more action issues we're good to release a new minor. Thanks much.

paulo-ferraz-oliveira:
Hm... it seems I can't merge since @starbelly has requested changes and these need to be approved (I even tried to remove him as reviewer to speed this up, but no good). We'll have to wait.

README.md
```diff
@@ -165,11 +165,30 @@ jobs:
...
           otp-version: '24'
-          gleam-version: '0.16'
-          rebar3-version: '3.16'
-      - run: rebar3 ct
+          gleam-version: '0.19.0-rc3'
+      - run: gleam test
+```
+
+### Example (Gleam on Ubuntu without OTP)
+
```
starbelly, line 173:
```suggestion
Note that `otp: false` currently is only applicable to gleam. 
```

starbelly: @paulo-ferraz-oliveira Thoughts?

michallepicki: I included this note below the example snippet. Should it be moved up?

starbelly: Ahh, nah that's good. I just didn't see, it should read fine in the actual README vs a PR :)

starbelly (approved):

starbelly:
approved

PR #82, opened 2022-01-10
Automate updating 3rd party licenses

### LINKED ISSUE #76: Build is failing per third-party license differences

paulo-ferraz-oliveira:
Does this seem like a good enough option?

Closes #76.

### DIFF
--- a/.github/workflows/update_3rd_party_licenses.sh
+++ b/.github/workflows/update_3rd_party_licenses.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -eux
+
+git config user.name "GitHub Actions"
+git config user.email "actions@user.noreply.github.com"
+git checkout main
+
+npm run licenses
+
+git add 3RD_PARTY_LICENSES
+git commit -m "Update 3rd party licenses" || true
+git push origin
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira:
I'm purposefully not updating the 3rd party licenses, to have this scheduled action happen on `main`, where it should work.

starbelly:
@paulo-ferraz-oliveira I was more thinking something that just runs a build, we should get notified if it fails. If it fails, then we know we need to update the licenses. 

The thing that I guess makes me nervous about this is nodejs deps 😬 Maybe I'm overly concerned here?

paulo-ferraz-oliveira:
Hm. I'm not worried, since the dependencies are kept by the action container itself. In any case, we can always try and see how it goes, right?

starbelly:
@paulo-ferraz-oliveira Fair enough. Just need to be sure to review commits before release :) 

Can you rebase this so it passes?

starbelly (approved):

paulo-ferraz-oliveira:
> Can you rebase this so it passes?

Sure. On it.

paulo-ferraz-oliveira:
Some action results are stuck because `ubuntu-16.04` is no longer available. I'm merging as-is and hope the merge of #83 will solve these issues in `main`.

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/actions/runs/1687446824 ran and didn't update anything. 😄 I assume it's because there's really nothing to update - we've updated the licenses very recently.

PR #84, opened 2022-01-10
Allow using rebar3 nightly builds

### LINKED ISSUE #68: Support non-tagged rebar3 versions?
As-is we're incapable of targeting the most recent non-tagged development commit, which could be useful for nightly CI as a way to preemptively detect upcoming issues. Should we do this? The only issue I haven't solved is "how" we'd fetch the version. We could go with a fixed string, eg. latest/main/... that we'd fetch if no tagged version was found (hopefully there'll never be a tag with the development branch's name).

An idea we might borrow from gleam-lang/setup-gleam is the use of `ref:main` as a way to "clone" from a ref and not fetch a given tagged element.

paulo-ferraz-oliveira:
Closes #68.


paulo-ferraz-oliveira:
The action itself will run `rebar3 version` and this might report an actual version (eg. at the moment, `3.18`) which could be confusing. Should we force the output to "nightly" when the input is so?

starbelly:
> The action itself will run `rebar3 version` and this might report an actual version (eg. at the moment, `3.18`) which could be confusing. Should we force the output to "nightly" when the input is so?

That makes sense to me 👍

paulo-ferraz-oliveira:
I made the changes and rebased on `main`.

paulo-ferraz-oliveira:
Fixing the `action` CI errors now...

starbelly (approved):

PR #85, opened 2022-01-10
Allow disabling problem matchers

### LINKED ISSUE #77: Allow disabling problem matchers
This is useful in 2 situations:

1. When you test against multiple Elixir/Erlang versions and you want to mark which one should produce the messages.
2. When you use alternative approach to producing the reports (for example [`mix_machine`](https://github.com/hauleth/mix_machine)) and you do not want to have duplicated messages.

This could be simple boolean flag `enable_problem_matchers` that would default to `true`, but it would allow power-users to disable such behaviour.

paulo-ferraz-oliveira:
Closes #77.

@hauleth, let me know if this change is good for you, as is. It's not `enable_`, since that was the default, but it serves the same purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ workflow by:
 - optionally, installing Gleam
 - optionally, installing `rebar3`
 - optionally, installing `hex`
-- Configures [problem matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) to show warnings and errors on PR's
+- optionally, having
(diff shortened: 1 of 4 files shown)


hauleth (approved):

starbelly (approved):

paulo-ferraz-oliveira:
This has been rebased on `main`.

paulo-ferraz-oliveira:
The checks that aren't passing are related to the use of `ubuntu-16.04`, solve in a different pull request. Squashing and merging...

PR #87, opened 2022-01-12
Fix failing CI

paulo-ferraz-oliveira:
Recent pull requests made it difficult to understand if actions were Ok or not; hopefully this is the last pull request before the next release.

~Also, there were some warnings regarding missing cert-related options when compiling Elixir. Let's see if these got solved remotely, also.~


paulo-ferraz-oliveira:
All ✅. No annotations.

starbelly (approved):

PR #92, opened 2022-02-16
Fix 3rd party license updater (automation)

paulo-ferraz-oliveira:
Instead of pushing directly, since the `main` branch is protected, this creates a pull request.

Failure: https://github.com/erlef/setup-beam/actions/runs/1852644324

It's possible this pull request doesn't pass CI since the 3rd party licenses are outdated, but once it's accepted+merged in at most 24h stuff should be back to normal.

### DIFF
--- a/.github/workflows/update_3rd_party_licenses.sh
+++ b/.github/workflows/update_3rd_party_licenses.sh
@@ -4,10 +4,25 @@ set -eux
 
 git config user.name "GitHub Actions"
 git config user.email "actions@user.noreply.github.com"
-git checkout main
 
+BRANCH=feature/3rd-party-licenses-update
+
+if git branch -a | grep "$BRANCH" > /dev/null; then
+    # already exists
+    exit
+fi
+
+git fetch origin
+git checkout -b "$BRANCH"
 npm run licenses
 
-git add 3RD_PARTY_LICENSES
-git commit -m "Update 3rd party licenses" || true
-git push origin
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
Here's the failing bit, as expected: https://github.com/erlef/setup-beam/runs/5224903868?check_suite_focus=true#step:14:12

starbelly:
@paulo-ferraz-oliveira It looks this is conflicting with https://github.com/erlef/setup-beam/blob/6c5caca0fb67184842b9de3a5fd39a1a1f51dea0/.github/workflows/action.yml#L31

paulo-ferraz-oliveira:
> @paulo-ferraz-oliveira It looks this is conflicting with
> 
> https://github.com/erlef/setup-beam/blob/6c5caca0fb67184842b9de3a5fd39a1a1f51dea0/.github/workflows/action.yml#L31

It's because the 3rd party licenses are outdated, which is what the pull requests wants to fix 😄 I'd prefer to not fix it here, merge, let the update happen via the action, and then "all's well" (the last time we fixed the files in the pull request, the action ran and failed after merge 😄). After that we can merge the other pull requests, following a rebase...

Let me know, though, if you prefer to do the fix here.

starbelly (approved):

paulo-ferraz-oliveira:
Thanks for the approval. I'm merging and running the action in main to see if the pull request gets created. 👍

PR #91, opened 2022-02-14
Allow for Windows'22 container

paulo-ferraz-oliveira:
⚠️ don't merge before #92 and the 3rd party licenses are update (lest we have "expected" errors which don't tell the whole story)

This pull request's goal is to close #90.

I've not tested this locally, so don't expect it to work from the get-go, but let's see...

PR #94, opened 2022-02-17
Add environment variables

### LINKED ISSUE #89: Dynamically discover include path on Windows
I'm using this action to build a NIF.

On Linux, the appropriate headers are automatically available to cmake. However, on Windows this isn't the case and a path has to be supplied manually to add to the include paths. Could this be exposed as some sort of variable to be referenced and passed into cmake?

The alternatives appear to be to either hard code the include path (but that's brittle and will break when a new version is released) or to include the header files locally (which also is brittle, but likely to be slower to break).

paulo-ferraz-oliveira:
⚠️ don't merge before #91+#92 and the 3rd party licenses are update (lest we have "expected" errors which don't tell the whole story)

This introduces some env. variables to the run, namely the ones identified in the `README.md`:
- `INSTALL_DIR_FOR_ELIXIR`: base folder for Erlang/OTP
- `INSTALL_DIR_FOR_GLEAM`: base folder for Elixir
- `INSTALL_DIR_FOR_OTP`: base folder for Gleam
- `INSTALL_DIR_FOR_REBAR3`: base folder for `rebar3`

and hopefully it closes #89 (still waiting on a comment/test by @samaaron)

### DIFF
--- a/dist/install-gleam.ps1
+++ b/dist/install-gleam.ps1
@@ -6,14 +6,16 @@ Set-Location $Env:RUNNER_TEMP
 
 $FILE_INPUT="gleam-${VSN}-windows-64bit.zip"
 $FILE_OUTPUT="gleam.zip"
-$DIR_FOR_BIN=".setup-beam/gleam/bin"
+$DIR_FOR_BIN=".setup-beam/gleam"
 
 $ProgressPreference="SilentlyContinue"
 Invoke-WebRequest "https://github.com/gleam-lang/gleam/releases/download/${VSN}/${FILE_INPUT}" -OutFile "$FILE_OUTPUT"
 $ProgressPreference="Continue"
-New-Item "$DIR_FOR_BIN" -ItemType Directory | Out-Null
+New-Item "${DIR_FOR_BIN}/bin" -ItemType Directory | Out-Null
 $ProgressPreference="SilentlyContinue"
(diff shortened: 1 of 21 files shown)


paulo-ferraz-oliveira:
I'm going to break this into 2 commits, since one of them is basically code maintenance and pollutes the revision.

dist/install-gleam.ps1
```diff
@@ -6,14 +6,16 @@ Set-Location $Env:RUNNER_TEMP
 
 $FILE_INPUT="gleam-${VSN}-windows-64bit.zip"
 $FILE_OUTPUT="gleam.zip"
-$DIR_FOR_BIN=".setup-beam/gleam/bin"
+$DIR_FOR_BIN=".setup-beam/gleam"
```
paulo-ferraz-oliveira, line 9: The tweaking of `bin`, here, was basically for consistency with the other outputs (otherwise we'd have to attach the `bin` to Gleam's base folder).

paulo-ferraz-oliveira:
70a59a5 basically adds a lot of `{` for consistency with existing code.

starbelly (approved):

paulo-ferraz-oliveira:
Rebased on top of `main`...

paulo-ferraz-oliveira:
This is now Ok, after the rebase. Merging...

Issue #99, opened 2022-03-31
Error installing rebar3


hermanius:
Using the following:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v3
    - name: Set up Elixir
      uses: erlef/setup-beam@v1
      with:
        elixir-version: '1.13.3' # Define the elixir version [required]
        otp-version: '24.1' # Define the OTP version [required]

I get the following error:
<img width="834" alt="image" src="https://user-images.githubusercontent.com/16594513/161158262-a13a5dd8-7058-46b4-8539-9e37917dddd9.png">

Am I doing something wrong?

hermanius:
The error stays, even if I add `install-rebar: false`

paulo-ferraz-oliveira:
Hm... let me test this.

paulo-ferraz-oliveira:
https://github.com/paulo-ferraz-oliveira/setup-beam-example/blob/setup-beam/issues/99/.github/workflows/ci.yml

Is this the same as your example? I don't see it error'ing out (maybe I'm missing something?)

Initially I thought it could be from the v3 (for the `checkouts` action), but seems not to be so.

paulo-ferraz-oliveira:
In particular, I have https://github.com/paulo-ferraz-oliveira/setup-beam-example/runs/5821121108?check_suite_focus=true#step:3:24, which seems to be different from your output.

hermanius:
hmmm, strange, I still have the same error. It has worked before, and suddenly stopped working, without any hanges in the yaml file. And the yaml files seems to look the same.
Is it an idea to give you temporarily access to the repository where I have the problem?

paulo-ferraz-oliveira:
I can fork it and test it out (if I know where stuff is), but I see no real reason for this to fail, atm.

Can you copy-paste from my example repo. and test again?

hermanius:
I tried the copy-paste, and only changed the on: push to on:workflow_dispatch, and I still get the same error.

I have send you an invite to collaborate on my private repository. It would be great if you have the time to check it out.

paulo-ferraz-oliveira:
> even if I add `install-rebar: false`

seems like an action issue, since it shouldn't try to install stuff we told it not to.

paulo-ferraz-oliveira:
[...]

PR #100, opened 2022-04-05
Handle strings as such

paulo-ferraz-oliveira:
This hopefully fixes one of the issues found via #99 (mis-handling of booleans).

**Note**: this'll most likely break stuff in-the-wild, since the way the values were being read (`getInput`) meant we'd always mix `rebar` and `hex`. We can either go with this approach (and bump major) or consider the default should've been `true`, from the start, and change that.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -27,7 +27,6 @@ jobs:
       - run: npm run shellcheck
       - run: npm run yamllint
       - run: npm run jslint
-      - run: npm run licenses
       - name: Check if build left artifacts
         run: git diff --exit-code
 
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira:
The values from the YAML aren't really booleans, though they _seem_ like it 😞

starbelly:
@paulo-ferraz-oliveira If I understand correctly, why not convert the strings to booleans on the way in so we don't have to do a major version?

paulo-ferraz-oliveira:
We can do that, too, I guess. Let me try it later. (I can't remember, by heart, how it would behave before the change, but can revert it all and try again)

paulo-ferraz-oliveira:
@starbelly, I'm trying your approach...

.
.
.

passed!

I'm also removing the license update, since we have this automated now, and working as expected (eg. #95, #97).

On the other hand I'm not merging just yet since there's a "Bad credentials" issue I've never dealt with 😕

starbelly (approved):

PR #102, opened 2022-04-11
Keep up with the times

paulo-ferraz-oliveira:
Just some house cleaning.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -43,5 +43,5 @@ outputs:
   rebar3-version:
     description: Exact version of rebar3 that was installed
 runs:
-  using: node12
+  using: node16
   main: dist/index.js
--- a/package.json
+++ b/package.json
@@ -19,24 +19,24 @@
   },
   "dependencies": {
     "@actions/core": "1.6.0",
-    "@actions/exec": "1.1.0",
-    "semver": "7.3.5"
+    "@actions/exec": "1.1.1",
+    "semver": "7.3.6"
   },
   "devDependencies": {
-    "@vercel/ncc": "0.33.1",
-    "eslint": "8.6.0",
-    "eslint-config-airbnb": "v19.0.4",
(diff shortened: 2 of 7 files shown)


action.yml
```diff
@@ -43,5 +43,5 @@ outputs:
   rebar3-version:
     description: Exact version of rebar3 that was installed
 runs:
-  using: node12
+  using: node16
```
paulo-ferraz-oliveira, line 46: 16 is now the default, I think since Nov last year.

package.json
```diff
@@ -19,24 +19,24 @@
...
+    "markdownlint": "0.25.1",
+    "prettier": "2.6.2",
     "shellcheck": "1.1.0",
     "yaml-lint": "1.2.4",
-    "yarn": "1.22.17"
+    "yarn": "1.22.18"
   },
   "engines": {
-    "node": ">=14"
+    "node": ">=16"
```
paulo-ferraz-oliveira, line 40: 16 is now the default, I think since Nov last year.

package.json
```diff
@@ -19,24 +19,24 @@
   },
```
paulo-ferraz-oliveira, line 19: These were bumped to their latest. None of them appears to have broken interface.

src/setup-beam.js
```diff
@@ -471,9 +471,11 @@ async function get(url0, pageIdxs) {
 
 function maybePrependWithV(versionToPrepend, specVersion) {
   const digitStart = /^\d+/
-  return digitStart.test(specVersion)
-    ? `v${versionToPrepend}`
-    : versionToPrepend
+  let v = versionToPrepend
```
paulo-ferraz-oliveira, line 474: I preferred the way it was, but I couldn't get the linter and formatter to work side-by-side. It didn't seem like a very bad compromise, this change.

starbelly (approved):

PR #110, opened 2022-04-22
Apply security fix as proposed by Dependabot

paulo-ferraz-oliveira:
I actually [went to the developer first](https://github.com/rasshofer/yaml-lint/issues/28#issuecomment-1106880768), pull requested the lib., but then [my update got rejected](https://github.com/rasshofer/yaml-lint/pull/30), since the lib. [moved to TypeScript (from JavaScript)](https://github.com/rasshofer/yaml-lint/issues/28#issuecomment-1106893662); still, a [new release](https://github.com/rasshofer/yaml-lint/releases/tag/v1.3.0) was done and this might just solve our issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
       },
       "devDependencies": {
         "@vercel/ncc": "0.33.4",
-        "eslint": "8.13.0",
+        "eslint": "8.14.0",
         "eslint-config-airbnb": "19.0.4",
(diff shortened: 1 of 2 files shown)


starbelly:
Ahh, so you got a patch done. I was wondering about that as I saw that it looked like we needed to upgrade to v2.

starbelly (approved):

PR #118, opened 2022-05-18
Test on top of Erlang/OTP 25+ (not RC)

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
This was done because of a test to an issue, a few minutes ago. In any case, it's an extra confidence boost 😄

dist/index.js
```diff
@@ -6193,7 +6193,7 @@ async function getOTPVersion(otpSpec0, osVersion) {
   const otpVersions = await getOTPVersions(osVersion)
   const otpSpec = otpSpec0.match(/^(OTP-)?([^ ]+)/)
   let otpVersion
-  if (otpSpec[1]) {
+  if (otpSpec[1] && !isStrictVersion()) {
```
paulo-ferraz-oliveira, line 6196: This was just us being super strict with input, but for the edge case being tested we do want to allow "Using -otp- or OTP-", even though it is to be made explicit via the version-type option.

starbelly (approved):

starbelly:
@paulo-ferraz-oliveira Approved, but left it open so you can squash as you like.

PR #134, opened 2022-08-22
Crash on invalid Elixir+OTP combo

### LINKED ISSUE #131: Non-crash on invalid Elixir+OTP combo
(initially reported by @tsloughter via Erlang Slack)

If we choose Elixir version 1.12 and Erlang version 25, the action (ran it today) will end up choosing Elixir 1.12.3 and Erlang 25.0.3. 

Because of a code validation, though, it'll not find `v1.12.3-otp-25` (from [builds.txt](https://repo.hex.pm/builds/elixir/builds.txt)) and will fall back to v1.12.3 (which is compiled with Erlang 22).

The question is: **does this validation actually make sense?** Or should we assume that if a combo is possible (for Elixir version+Erlang major) it'll always be found in `builds.txt` and that the fallback can be removed from our code?

Adding @ericmj @wojtekmach to recruit your help. Thanks much.

P.S.: the bit of code in question is this https://github.com/erlef/setup-beam/blob/main/src/setup-beam.js#L200.

paulo-ferraz-oliveira:
Closes #131.

**Edit**: this also fixes some tests that started failing due to our new pre-conditions.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -116,6 +116,7 @@ jobs:
           - elixir-version: 'master'
             otp-version: '25'
             os:  'ubuntu-20.04'
+            version-type: 'strict'
           - gleam-version: 'nightly'
             otp-version: '24'
           - gleam-version: '0.23'
--- a/__tests__/setup-beam.test.js
+++ b/__tests__/setup-beam.test.js
@@ -171,8 +171,8 @@ async function testElixirVersions() {
   let otpVersion
 
   spec = '1.1.x'
-  otpVersion = 'OTP-23'
-  expected = 'v1.1.1'
+  otpVersion = 'OTP-17'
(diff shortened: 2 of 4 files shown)


wojtekmach (approved):
👍

paulo-ferraz-oliveira:
@wojtekmach, https://github.com/erlef/setup-beam/runs/7950029010?check_suite_focus=true#step:3:15 is failing because it can't find `master+25`. Should we open an exception, in the code, for `master`? I see that `-otp-25` only shows up some times, so not sure what the rule is, here.

paulo-ferraz-oliveira:
> I see that -otp-25 only shows up some times, so not sure what the rule is, here.

Upon further inspection, I see it shows up whenever there's a compatible Elixir/Erlang combo. I'll open an exception to `master` later, before I merge this pull request in...

PR #138, opened 2022-09-02
Update for latest Gleam

lpil:
- Gleam releases now use LLVM triplets for platforms so the download
  URLs have been updated accordingly.
- Gleam's rebar3 support has been removed.

This change means that future versions of this action cannot download
older versions of Gleam. I can make it backwards compatible, but I will
need help from someone who can convert the shell script into PowerShell
for Windows.


paulo-ferraz-oliveira:
Hi, @lpil.

I'd usually not worry about backward compatibility before 1.0.0 (following Semantic Versioning principles) - which is Gleam's case.

At the same time, if `rebar3`'s no longer support, would it be better to remove `rebar3-version` from the Gleam combos, everywhere?

Cheers.

starbelly:
> This change means that future versions of this action cannot download
> older versions of Gleam. I can make it backwards compatible, but I will
> need help from someone who can convert the shell script into PowerShell
> for Windows.

This is an interesting one... does a backward incompatible change for a supported language mean that we break compatibility for users and thus need to bump the major version? It may be the a big note in the README.md is good enough. 

I'm not 100% sure on this on tbh. I tend to think that since our API has not changed, then we do not need to bump the major. To  boot, gleam is not 1.0 yet, thus it should be somewhat expected, no? 

Thoughts?

lpil:
I already have the code for backwards compatibility using shell, so I could add that and we leave windows support as a future addition?

starbelly:
> I already have the code for backwards compatibility using shell, so I could add that and we leave windows support as a future addition?

That sounds quite amenable to me :)

paulo-ferraz-oliveira:
@lpil, sure thing. Would you create an issue for what would need to be done, so somebody can take a look at it later? Cheers.

lpil:
Fab! I've updated this PR and created the issue.

The build is failing but isn't printing any debug information. Is there a way to get more out of it? It passes for other installations so I'm not sure what might be happening.

starbelly:
> Fab! I've updated this PR and created the issue.
> 
> The build is failing but isn't printing any debug information. Is there a way to get more out of it? It passes for other installations so I'm not sure what might be happening.

There's no gleam-0.23.0-rc1-linux-amd64.tar.gz for [v0.23.0-rc1](https://github.com/gleam-lang/gleam/releases/tag/v0.23.0-rc1)

lpil:
Ah there it is! I found a typo in the script.

starbelly (approved):

starbelly:
@paulo-ferraz-oliveira LGTM.

lpil:
Oh yes! I forgot to remove the rebar bits. I have done that now

paulo-ferraz-oliveira:
Merging... Many thanks, @lpil.

lpil:
Wonderful, thank you! Would it be possible to get this released soon? I would like to start using new Gleam versions 😅

starbelly:
> Oh yes! I forgot to remove the rebar bits. I have done that now

Yup, we certainly can 👍

paulo-ferraz-oliveira:
I'll release and tag this today, and by then it'll also contain a pull request of mine 😄.

paulo-ferraz-oliveira:
@starbelly, I'll move your tag, if that's Ok, since it was never released in any case...

paulo-ferraz-oliveira:
v1.13.0 released, v1.13 tagged, v1 moved (to v1.13)

PR #139, opened 2022-09-07
Install OTP on Windows to a local directory

wojtekmach:
It is especially useful not to install things globally on self-hosted runner as these may not be running in isolation.

By installing OTP into a known location we could remove the code that was checking where exactly it was installed.

This PR also removes the `echo $INSTALL_DIR_FOR_OTP >> $GITHUB_ENV` (and the PowerShell equivalent) as it never worked. The variable name should have been `$GITHUB_PATH`, not `$GITHUB_ENV`. Again, no longer necessary as we install OTP to a known location.

### DIFF
--- a/dist/install-otp.sh
+++ b/dist/install-otp.sh
@@ -17,4 +17,4 @@ tar zxf "${FILE_OUTPUT}" -C "${DIR_FOR_BIN}" --strip-components=1
 echo "Installed Erlang/OTP version follows"
 ${DIR_FOR_BIN}/bin/erl -version
 
-echo "INSTALL_DIR_FOR_OTP=${RUNNER_TEMP}/${DIR_FOR_BIN}" >> "${GITHUB_ENV}"
+echo "INSTALL_DIR_FOR_OTP=${RUNNER_TEMP}/${DIR_FOR_BIN}"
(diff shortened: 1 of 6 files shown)


paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
Good stuff. Approved. Thanks.

dist/install-otp.sh
```diff
@@ -17,4 +17,4 @@ tar zxf "${FILE_OUTPUT}" -C "${DIR_FOR_BIN}" --strip-components=1
 echo "Installed Erlang/OTP version follows"
 ${DIR_FOR_BIN}/bin/erl -version
 
-echo "INSTALL_DIR_FOR_OTP=${RUNNER_TEMP}/${DIR_FOR_BIN}" >> "${GITHUB_ENV}"
```
wojtekmach, line 20: This change was a mistake as it stopped setting the env var which was done via writing to GITHUB_ENV file. Ill try to send a fix tomorrow. Sorry about that.

paulo-ferraz-oliveira: No prob. I should've reviewed better... and also tests should've probably failed (unless that's not important for tests).

PR #141, opened 2022-09-15
Fix setting INSTALL_DIR_FOR_OTP

wojtekmach:
Ref: https://github.com/erlef/setup-beam/pull/139#discussion_r970967071

Before: (See "List environment variables" step)
- Ubuntu: https://github.com/erlef/setup-beam/actions/runs/3043092234/jobs/4901979442
- Windows: https://github.com/erlef/setup-beam/actions/runs/3043092227/jobs/4901979281

After:

- Ubuntu: https://github.com/erlef/setup-beam/actions/runs/3059580103/jobs/4937122642
- Windows: https://github.com/erlef/setup-beam/actions/runs/3059580089/jobs/4937122295

### DIFF
--- a/dist/install-otp.ps1
+++ b/dist/install-otp.ps1
@@ -14,4 +14,4 @@ Start-Process "${FILE_OUTPUT}" "/S /D=${ERL_ROOT}" -Wait
 Write-Output "Installed Erlang/OTP version follows"
 & "${ERL_ROOT}/bin/erl.exe" "+V" | Write-Output
 
-"INSTALL_DIR_FOR_OTP=" + ${ERL_ROOT}
+"INSTALL_DIR_FOR_OTP=${ERL_ROOT}" | Out-File -FilePath ${Env:GITHUB_ENV} -Encoding utf8 -Append
--- a/dist/install-otp.sh
+++ b/dist/install-otp.sh
@@ -17,4 +17,4 @@ tar zxf "${FILE_OUTPUT}" -C "${DIR_FOR_BIN}" --strip-components=1
 echo "Installed Erlang/OTP version follows"
 ${DIR_FOR_BIN}/bin/erl -version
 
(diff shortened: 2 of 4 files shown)


wojtekmach:
On Windows we have this env vars output:

```
D:\a\_temp/.setup-beam/elixir
D:\a\_temp/.setup-beam/gleam
D:\a\_temp\.setup-beam\otp
D:\a\_temp/.setup-beam/rebar3
```

notice it's mix of `/` and `\` everywhere besides now OTP. Happy to make OTP like the others or change others to use just `\`. Happy to do it here or in a separate PR.

paulo-ferraz-oliveira:
> notice it's mix of / and \ everywhere besides now OTP. Happy to make OTP like the others or change others to use just \. Happy to do it here or in a separate PR.

If `\` is more like Windows users are used to, go for it 😄. I used it (`/`), during development, to ease diff.s between the *nix and Windows scripts, which is why you'll find similar variable names, and whatnot. (but the scripts didn't grow much, so that's probably manageable, in any case).

On the other hand, should we add some simple tests to this to make sure the env. variables are set?

I'm approving, in any case, to ease your merge, since I believe that it's good to ship even as-is.

paulo-ferraz-oliveira (approved):

wojtekmach:
Agreed about testing this. We already have a job that prints these env vars so I think an easy improvement is this:

```diff
- echo $A
+ [ -n "$A" ] && echo $A
```

I'll send a patch soon!

paulo-ferraz-oliveira:
Good stuff, @wojtekmach.

paulo-ferraz-oliveira:
I'll release this as 1.13.1. The next PR can go later...

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/releases/tag/v1.13.1
v1 and v13 moved.

wojtekmach:
Thanks!

PR #144, opened 2022-10-02
Allow version-type: strict + otp-version: master for Elixir -based builds

paulo-ferraz-oliveira:
(**edit**: though the title is somewhat restrictive I believe this fix should work for Gleam -based builds in the same way)

@ericmj found that we can't use combo

```
  otp-version: master
  elixir-version: 1.14.0-otp-master
```

but this is unintentional, as we should, with `version-type: strict`. This pull request aims to fix that.


paulo-ferraz-oliveira:
We add a workflow -based test and a unit test for this change. Since I'm doing only additions to the code, I'm pretty confident there's no regressions (believing that all previous tests were already very comprehensive).

paulo-ferraz-oliveira:
Ah, wait, spoke too soon. This appears to have some kind of influence of previously working tests. Now I just need to make sure the tests were meaningful and correct as per this change.

paulo-ferraz-oliveira:
I guess `strict` still tries to do some guess work in unexpected cases. I'll fix that.

paulo-ferraz-oliveira:
@ericmj, what should happen if a developer chooses Elixir `master` but OTP `25`? As opposed to Elixir `master` and OTP `24`? In the latter we know how to choose `master-otp-24`, but how should we know to only choose `master` if OTP is `25`?

.github/workflows/ubuntu.yml
```diff
@@ -21,6 +21,13 @@ jobs:
       fail-fast: false
       matrix:
         combo:
+          - otp-version: 'master'
```
paulo-ferraz-oliveira, line 24: Added as part of the "was failing, but shouldn't!" testbed.

.github/workflows/ubuntu.yml
```diff
@@ -113,12 +120,13 @@ jobs:
           - elixir-version: 'master'
             otp-version: '23.1'
             os:  'ubuntu-20.04'
-          - elixir-version: 'master'
+          - elixir-version: 'main'
```
paulo-ferraz-oliveira, line 123: This was actually a fix. Since OTP 25, Elixir no longer has `master`, but `main`.

.github/workflows/ubuntu.yml
```diff
@@ -113,12 +120,13 @@ jobs:
...
             otp-version: '25'
             os:  'ubuntu-20.04'
             version-type: 'strict'
           - gleam-version: 'nightly'
             otp-version: '24'
+            os: 'ubuntu-latest'
```
paulo-ferraz-oliveira, line 129: I believe this should've been broken, but can't understand how we didn't see it before. In any case, on the README, the Ubuntu build is broken and there were recently issues with running actions on GitHub actions [like so](https://github.com/erlef/setup-beam/actions/runs/3066558491), which was hiding bugs.

__tests__/setup-beam.test.js
```diff
@@ -205,9 +213,9 @@ async function testElixirVersions() {
   simulateInput('version-type', 'loose')
 
   simulateInput('version-type', 'strict')
-  spec = 'master'
+  spec = 'main'
```
paulo-ferraz-oliveira, line 216: Explained above, though probably only meaningful for OTP 25+

dist/index.js
```diff
@@ -7189,8 +7189,8 @@ async function main() {
   const rebar3Spec = core.getInput('rebar3-version', { required: false })
 
   if (otpSpec !== 'false') {
-    const otpVersion = await installOTP(otpSpec, osVersion)
```
paulo-ferraz-oliveira, line 7192: We install Elixir based on the direct OTP input, and not the output of this function (easier to reason about)...

dist/index.js
```diff
@@ -7300,19 +7300,13 @@ async function maybeInstallRebar3(rebar3Spec) {
...
-      `Requested Erlang/OTP version (${otpSpec0}) ` +
-        "should not contain 'OTP-, or maint-'",
-    )
-  }
-  if (otpSpec) {
-    otpVersion = getVersionFromSpec(
-      otpSpec[2],
-      Array.from(otpVersions.keys()).sort(),
-    )
+  let otpSpec = otpSpec0 // might be a branch (?)
```
paulo-ferraz-oliveira, line 7303: Greatly simplified. We drop OTP-, maint- and whatnot to favour "branch over version," as suggested by @ericmj.

dist/index.js
```diff
@@ -7324,60 +7318,46 @@ async function getOTPVersion(otpSpec0, osVersion) {
   return otpVersions.get(otpVersion) // from the reference, for download
 }
 
-async function getElixirVersion(exSpec0, otpVersion) {
+async function getElixirVersion(exSpec0, otpVersion0) {
```
paulo-ferraz-oliveira, line 7321: Also greatly simplified as per previous explanation.

dist/index.js
```diff
@@ -7438,7 +7420,6 @@ async function getOTPVersions(osVersion) {
         })
     })
   }
-
```
paulo-ferraz-oliveira, line 7441: Unintended...

paulo-ferraz-oliveira:
@ericmj, did you get around to testing this?

paulo-ferraz-oliveira:
@starbelly, I'm recruiting your help to validate these changes, as @ericmj might not be available at this moment. They are not only pertinent to the initial problem issued by @ericmj, but also generally to improve the code base. I can walk you through the changes if there's something you don't understand or need further detail on.

starbelly:
The changes look reasonable, let me give this a whirl in a test project tomorrow.

dist/index.js
starbelly: Nice one!

src/setup-beam.js
```diff
@@ -154,60 +148,46 @@ async function getOTPVersion(otpSpec0, osVersion) {
...
-
-  const exSpec = exSpec0.match(/^v?(.+)(-otp-.+)/) || exSpec0.match(/^v?(.+)/)
-  let elixirVersion
-  if (exSpec[2] && !isStrictVersion()) {
-    throw new Error(
-      `Requested Elixir / Erlang/OTP version (${exSpec0} / ${otpVersion}) ` +
-        "should not contain '-otp-...'",
-    )
+  const exSpec = exSpec0.replace(/-otp-.*$/, '')
+  const elixirVersionFromSpec = getVersionFromSpec(exSpec, semverVersions, true)
```
starbelly, line 157: This is the only thing I don't like, the boolean blindness, but it's javascript so 🤷 😁  The only thing I could thing to do improve this is to pass around an object, I would not block on it though.

starbelly (approved):
@paulo-ferraz-oliveira This looks solid to me. Tested and all.  Since it's not a small change, we may want to hold off on force pushing it the v1 tag and maybe ask some other to try it for a bit. I'll certainly give it a solid whirl at work for a while. What do you think? 

Regardless, approved 🎯

ericmj (approved):
Thank you @paulo-ferraz-oliveira. This works perfectly https://github.com/hexpm/hex/actions/runs/3284596299.

paulo-ferraz-oliveira:
@starbelly, I can push the minor and patch only, but will need a comment from you, or anybody else, in a few days, to force-push `v1`.

paulo-ferraz-oliveira:
v1.14 and v1.14.0 pushed. I'll create the release now. @starbelly, do let us know soonish, so we can force push `v1`. Thanks.

starbelly:
@paulo-ferraz-oliveira Ship it!! 🥳

PR #153, opened 2022-11-08
Ease Elixir escripts execution by adding ${HOME}/.mix/escripts to path

btkostner:
This adds the mix escripts folder to `PATH`, allowing you to run `mix escript.install hex protobuf --force` and have access to the protobuf generator right away. Making life simpler.

I'm having this as a WIP because I don't program in windows and need to confirm this actually works as intended.

### DIFF
--- a/src/install-elixir.ps1
+++ b/src/install-elixir.ps1
@@ -18,4 +18,7 @@ $ProgressPreference="Continue"
 Write-Output "Installed Elixir version follows"
 & "${DIR_FOR_BIN}/bin/elixir" "-v" | Write-Output
 
+$ProgressPreference="Continue"
+New-Item "%UserProfile%\.mix\escripts" -ItemType Directory | Out-Null
+
 "INSTALL_DIR_FOR_ELIXIR=${Env:RUNNER_TEMP}/${DIR_FOR_BIN}" | Out-File -FilePath ${Env:GITHUB_ENV} -Encoding utf8 -Append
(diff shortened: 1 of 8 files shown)


paulo-ferraz-oliveira:
Hey, @btkostner. Thanks for this.

I'm wondering if this is sufficiently generic though, to have included in the action.

What do you think, @starbelly, @ericmj, @wojtekmach?

wojtekmach:
+1 from me.

src/install-elixir.ps1
```diff
@@ -18,4 +18,7 @@ $ProgressPreference="Continue"
 Write-Output "Installed Elixir version follows"
 & "${DIR_FOR_BIN}/bin/elixir" "-v" | Write-Output
 
+$ProgressPreference="Continue"
+New-Item "%UserProfile%\.mix\escripts" -ItemType Directory | Out-Null
```
paulo-ferraz-oliveira, line 22:
```suggestion
New-Item "%UserProfile%/.mix/escripts" -ItemType Directory | Out-Null
```

paulo-ferraz-oliveira: This works and reduces the number of differences to the Linux script we use to have similar behaviour.

paulo-ferraz-oliveira:
@btkostner, could you add a simple test that demonstrates the behaviour, and can be constantly used for regression testing?

btkostner:
Alright, I added it to the existing install test matrix, but if you want me to split it off to it's own job, I can do that. Tests show it working in Linux and Windows. Let me know if there is more that I should do.

paulo-ferraz-oliveira (approved):
This looks good. Thanks @btkostner.

paulo-ferraz-oliveira:
Merging to `main`...

Issue #155, opened 2022-11-14
Strange build issues on GHA


lucacorti:
Hi,

Builds for a number of projects I have on github using setup-beam started failing suddenly in a very consistent way despite I made no changes to them. This is an example:

https://github.com/lucacorti/jsonapi_plug/actions/runs/3460577293/jobs/5777255548

It's interesting they are only failing in the test step (matrix based with multiple versions), while the lint step (matrix with only one combination) seems to work fine.

paulo-ferraz-oliveira:
Hi, @lucacorti.

We fetch our version lists from here (e.g. for Ubuntu 22.04): https://repo.hex.pm/builds/otp/ubuntu-22.04/builds.txt.

One of the errors you show

```
Requested Erlang/OTP version (23) not found in version list (should you be using option 'version-type': 'strict'?)
```

is indeed correct according to that list and this [action's README](https://github.com/erlef/setup-beam#compatibility-between-operating-system-and-erlangotp).

You seem to be targeting `ubuntu-latest` which might have changed recently, thus causing that issue (I will validate soon). With that strategy this might happen again in the future given `-latest` will always be a running target.

lucacorti:
Oh I see, that makes sense indeed. I'll try setting 20.04 as the target.

Thanks!

paulo-ferraz-oliveira:
[GitHub Actions: Ubuntu-latest workflows will use Ubuntu-22.04](https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/) was five days ago (November 9, 2022). Does this coincide with your "failing suddenly"? If it does, feel free to close the issue, or keep it open if you want to discuss something further.

lucacorti:
Yes, that's it. I failed to notice the OS comaptibility bit. Thanks!

PR #156, opened 2022-11-14
New option hexpm-mirrors, to allow specifying several mirrors and change default repo.hex.pm

pguyot:
New option hexpm-mirrors to specify the mirror(s) to use (defaults to repo.hex.pm). This is particularly useful in cases `repo.hex.pm` is not reliable as it proved to be recently.

Signed-off-by: Paul Guyot <pguyot@kallisys.net>


.github/workflows/test-cache-hexpm.yml
```diff
@@ -0,0 +1,34 @@
...
+      - main
+
+jobs:
+  cache_hexpm:
+    name: Cache Hex.pm
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'windows-latest']
```
paulo-ferraz-oliveira, line 19: This will eventually fail once we reach a point where OTP 24 is not supported for `-latest`. I'd prefer a fixed version, that we can later revisit.

paulo-ferraz-oliveira:
I'm Ok-ish with this change, though I feel that the action shouldn't care about cache. Specifically, you're doing this for Hex (Elixir), but e.g. not rebar3 (Erlang), and if we go down that road we might end up with something outside of our original scope. I believe we've discussed this with other maintainers, in the past (@starbelly, @ericmj, @wojtekmach) and ended up not including it. I'll search the issues and pull requests.

paulo-ferraz-oliveira:
From https://github.com/erlef/setup-beam/pull/45 I can find https://gist.github.com/eproxus/d74315864fd6897bb47741e8de5bc980; and we ended up not including cache-related instructions in the README.

paulo-ferraz-oliveira:
I'm also thinking that somebody comes along and wants to add cache for `rebar3`'s PLT management; then somebody feels we should have cache for `rustler`... you can see where this is going 😄

pguyot:
The main issue here is that `erlef/setup-beam` action uses hex.pm a lot (for erlang, elixir and mix packages), and when a matrix strategy is big enough, network errors occur, making the action totally unusable. It's not about performance, it's about getting a consistent CI. There is no point in caching rebar as it's provided by GitHub/AWS S3.

These failures represent most if not all recent failures on the master branch for [AtomVM](/atomvm/AtomVM) project and a lot of failures on PRs
https://github.com/atomvm/AtomVM/actions/workflows/build-and-test.yaml?query=branch%3Amaster+is%3Afailure
https://github.com/atomvm/AtomVM/actions/workflows/build-and-test-other.yaml?query=branch%3Amaster+is%3Afailure

OTOH, setup-* actions definitely should handle cache, the official ones are automated cache providers for packages.
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

So maybe here the name cache-hexpm is misleading as it's about caching downloads, not built items (although arguably, it caches built mix archives for rebar & hex).

paulo-ferraz-oliveira:
> making the action totally unusable

Are these CDN -related? What types of issues are we talking about? (because the assets we use are behind a CDN, already)

> There is no point in caching rebar as it's provided by GitHub/AWS S3

I meant "`rebar3` dep.s". Though many are provided (and potentially used from Hex) not all are. But apparently the scope of the pull request is different from what the title an option name suggest.

...

OTOH if what you're providing is caching of the action elements itself, I'm more receptive. I didn't look at the details of the implementation (it's late; this wasn't discussed before, ...) but maybe I was mislead exactly because of the option name. In that case, why would we need an option for this? If cache is working and makes building faster and more consistent, shouldn't it simply be integrated into the action?

paulo-ferraz-oliveira:
> (although arguably, it caches built mix archives for rebar & hex).

If it does, could we separate these from the previous ones? The action's elements themselves (?)

pguyot:
I double-checked and actually, this caches the unzipped hex.ez.
Previous code didn't cache properly rebar from hex.pm as downloaded by mix, it now does.

hex.pm is supposedly a CDN, but there have been a lot of issues recently.

paulo-ferraz-oliveira:
Great, thanks. I'll wait to hear from others until we decide how to move forward (and then do a proper review).

pguyot:
Drawing from this discussion, let's consider an alternative implementation.
A new option to choose the mirror to use to download the action's assets (erlang for Linux, elixir, rebar) and cache this on GitHub (by default).

paulo-ferraz-oliveira:
My question still holds: "... why would we need an option for this? If cache is working [from your PR onward] and makes building faster and more consistent, shouldn't it simply be integrated into the action?" meaning... why does your new behaviour simply not become the default? What are we expected to solve with an option?

src/setup-beam.js
```diff
@@ -17,21 +18,50 @@ async function main() {
...
 
   if (otpSpec !== 'false') {
-    await installOTP(otpSpec, osVersion)
-    const elixirInstalled = await maybeInstallElixir(elixirSpec, otpSpec)
+    const otpVersion = await getOTPVersion(otpSpec, osVersion)
+    const elixirVersion = await getElixirVersion(elixirSpec, otpVersion)
+    // No cache for Windows OTP as it's downloaded from GitHub
+    const useCache =
+      hexPmMirror.endsWith(';cache') &&
+      otpVersion !== 'master' &&
```
ericmj, line 32: This needs to check for all branches and also check branches for Elixir.

ericmj:
> The main issue here is that erlef/setup-beam action uses hex.pm a lot (for erlang, elixir and mix packages), and when a matrix strategy is big enough, network errors occur, making the action totally unusable. It's not about performance, it's about getting a consistent CI. There is no point in caching rebar as it's provided by GitHub/AWS S3.

This is strange to me because repo.hex.pm and github assets use the same CDN and will hit the same POPs/data centers.

Are you using github runners or your own? I haven't heard any other reports of github runners having network issues to repo.hex.pm. I also cannot see any substantial amount of errors in our metrics. Are you having more success using mirrors?

pguyot:
> > The main issue here is that erlef/setup-beam action uses hex.pm a lot (for erlang, elixir and mix packages), and when a matrix strategy is big enough, network errors occur, making the action totally unusable. It's not about performance, it's about getting a consistent CI. There is no point in caching rebar as it's provided by GitHub/AWS S3.
> 
> This is strange to me because repo.hex.pm and github assets use the same CDN and will hit the same POPs/data centers.

Yet it does happen when the action matrix is large enough.

> Are you using github runners or your own? I haven't heard any other reports of github runners having network issues to repo.hex.pm. I also cannot see any substantial amount of errors in our metrics. Are you having more success using mirrors?

This is GitHub runners.

I even had the problem running the CI to test my changes. See for example:
https://github.com/pguyot/setup-beam/actions/runs/3465494731
Or just the test that was run for this very PR:
https://github.com/erlef/setup-beam/actions/runs/3466560021/jobs/5796596347

I didn't have any problem with [a](https://cdn.jsdelivr.net/hex) mirror, but I didn't test it sufficiently. My plan was to use default CDN with GH cache.

paulo-ferraz-oliveira:
@pguyot, have you tested the result of this change, with your (currently failing) workflow, to make sure it solves the issues you want it to? (you can replace `erlef/setup-beam` with `pguyot:w46/add-cache-for-hex.pm`)

Can you maybe post the results here, so we can check the difference?

Thanks.

pguyot:
> My question still holds: "... why would we need an option for this? If cache is working [from your PR onward] and makes building faster and more consistent, shouldn't it simply be integrated into the action?" meaning... why does your new behaviour simply not become the default? What are we expected to solve with an option?

I am confused. The PR code now makes caching of `repo.hex.pm` the default, I did amend the code from your review.

paulo-ferraz-oliveira:
> I am confused. The PR code now makes caching of repo.hex.pm the default, I did amend the code from your review.

You're right; my text was probably not super explicit, but I did ask:

* ... why would we need **an option** for this? 
* What are we expected to solve with **an option**?

i.e. why would https://github.com/erlef/setup-beam/pull/156/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R36 be required as an option. Do you foresee a case where not using the cache, as you propose, would be better?

pguyot:
> i.e. why would https://github.com/erlef/setup-beam/pull/156/files#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R36 be required as an option. Do you foresee a case where not using the cache, as you propose, would be better?

This uses some space in the 10GB that GitHub provides for caching. About 100 MB. That's the reason I considered parsing ";cache" at the end of the URL. But this could be removed and the logic simplified if you think it's better, the complexity might indeed not be worth it.

paulo-ferraz-oliveira:
Yeah, I believe if we move forward (still waiting on @starbelly and @wojtekmach to say something - hopefully they will, if they have time; don't know if @ericmj has anything else to say) we should remove the option and simplify the change as much as possible. 100 MB, out of the 10 GB, is acceptable, I guess.

wojtekmach:
I don't recall previous conversations around this but in my opinion built-in support for caching seems like a good idea, I'd expect most people would like to use it most of the time. I believe I'm caching deps on most if not all of my projects. I'd even say cache "on" by default is a good idea but we need to consider drawbacks. Are there any?

> (...) add cache for rebar3's PLT management; then somebody feels we should have cache for rustler.

This is a very fair concern. I'd say the action should be concerned with caching only the things it downloads by itself. So I think we'd cache rebar but not rebar PLTs, one would need to do that themselves, which is not complicated.

I'm not sure about coupling caching with mirroring though, I think these are orthogonal concerns. For example, if we'd ever want to cache the `gleam` executable, we'd most likely download it from somewhere else than repo.hex.pm.

paulo-ferraz-oliveira:
I think we all agree that the action could/should cache the elements it depends on, but stuff like PLTs, consumer-specific dependencies, etc. should not be included (which is what I thought this pull request was for, initially). Is it possible to separate these, @pguyot? Or is it a consequence of your implementation that the Hex cache folder is consumed wholly.

pguyot:
> I think we all agree that the action could/should cache the elements it depends on, but stuff like PLTs, consumer-specific dependencies, etc. should not be included (which is what I thought this pull request was for, initially). Is it possible to separate these, @pguyot? Or is it a consequence of your implementation that the Hex cache folder is consumed wholly.

It does not and never did cache stuff like PLTs and consumer-specific dependencies.
The latest amended version removes the cache option and simplifies as discussed here.

.github/workflows/test-hexpm-mirrors.yml
```diff
@@ -0,0 +1,35 @@
...
+      - main
+
+jobs:
+  test_hexpm_mirror_option:
+    name: Usage of hexpm-mirror option
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'windows-latest']
```
paulo-ferraz-oliveira, line 19: I'd use something "not latest" here. Otherwise, when the target finally moves, in a few months/years, it is possible the tests start failing, as explained in [this comment](https://github.com/erlef/setup-beam/issues/155#issuecomment-1314502264).

.github/workflows/test-hexpm-mirrors.yml
```diff
@@ -0,0 +1,35 @@
+---
```
paulo-ferraz-oliveira, line 1: Let's rename the file `hexpm-mirror.yml`, and `name`, below, too.

.github/workflows/test-hexpm-mirrors.yml
```diff
@@ -0,0 +1,35 @@
...
+name: test-hexpm-mirror
```
paulo-ferraz-oliveira, line 2:
```suggestion
name: hexpm-mirror
```

.github/workflows/test-hexpm-mirrors.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test_hexpm_mirror_option:
```
paulo-ferraz-oliveira, line 13:
```suggestion
  test_option_hexpm_mirror:
```

.github/workflows/test-hexpm-mirrors.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: Usage of hexpm-mirror option
```
paulo-ferraz-oliveira, line 14:
```suggestion
    name: Test option hexpm-mirror
```

src/install-elixir.ps1
```diff
@@ -4,12 +4,14 @@ $ErrorActionPreference="Stop"
 
 Set-Location ${Env:RUNNER_TEMP}
 
+$CACHEDIR="${Env:RUNNER_TEMP}\.setup-beam\cache"
```
paulo-ferraz-oliveira, line 7:
```suggestion
$CACHEDIR="${Env:RUNNER_TEMP}/.setup-beam/cache"
```

src/install-elixir.ps1
```diff
@@ -4,12 +4,14 @@ $ErrorActionPreference="Stop"
...
 $FILE_INPUT="${VSN}.zip"
-$FILE_OUTPUT="elixir.zip"
+$FILE_OUTPUT="${CACHEDIR}\elixir-${VSN}.zip"
```
paulo-ferraz-oliveira, line 9:
```suggestion
$FILE_OUTPUT="${CACHEDIR}/elixir-${VSN}.zip"
```

src/install-elixir.sh
```diff
@@ -5,11 +5,13 @@ set -eo pipefail
 cd "${RUNNER_TEMP}"
 
 VSN=${1}
+CACHEDIR="${RUNNER_TEMP}/.setup-beam/cache"
 FILE_INPUT="${VSN}.zip"
-FILE_OUTPUT=elixir.zip
+FILE_OUTPUT="${CACHEDIR}/elixir-${FILE_INPUT}"
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira:
```suggestion
FILE_OUTPUT="${CACHEDIR}/elixir-${VSN}.zip"
```

src/install-otp.ps1
```diff
@@ -4,7 +4,7 @@ $ErrorActionPreference="Stop"
 
 Set-Location ${Env:RUNNER_TEMP}
 
-$FILE_OUTPUT="otp.exe"
+$FILE_OUTPUT="otp_win64_${VSN}.exe"
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: It also seems we're missing `CACHEDIR`, and friends, in this file.

src/setup-beam.js
```diff
@@ -149,6 +174,9 @@ async function getOTPVersion(otpSpec0, osVersion) {
 }
 
 async function getElixirVersion(exSpec0, otpVersion0) {
+  if (!exSpec0) {
+    return undefined
```
paulo-ferraz-oliveira, line 178: Could you add a comment explaining why this is required?

paulo-ferraz-oliveira:
@pguyot, this is looking better to me. I started the review before, but only finished it now. After you apply the requested changes I'll still compare the `sh` and `ps1` files and potentially request (suggest) further changes.

Also, it seems one of the tests is failing.

src/install-otp.ps1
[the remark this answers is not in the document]
pguyot: This is on purpose: there is no cache for Windows OTP because it is downloaded from GitHub, not from hex.pm.

paulo-ferraz-oliveira:
> Yes, this is flappy

I understand this might happen in consumers of the action, but the action tests should never fail; the knowledge of "having to repeat because it failed" will be lost in time. Test results should be repeatable and 🟢 all the time.

Also, what I see is "https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5860609846#step:3:26" which would further be cryptic, in future updates, regarding "tests might sometimes fail", since these are not timeouts...

paulo-ferraz-oliveira:
> because the entry was also cached in parallel from repo.hex.pm

If a job fails the cache is not created. I assume here you're saying that those that passed will create cache, and the failing ones will hit the CDN less. Is that it?

pguyot:
> > because the entry was also cached in parallel from repo.hex.pm
> 
> If a job fails the cache is not created. I assume here you're saying that those that passed will create cache, and the failing ones will hit the CDN less. Is that it?

The cache key depends on various parameters and in the current test file, there is at least one other test case with the same cache key (using repo.hex.pm mirror). It passed and therefore it created the entry. So if you just re-run this test, it will pass for sure.

.github/workflows/hexpm-mirror.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  test_option_hexpm_mirror:
+    name: Test option hexpm-mirror
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-22.04', 'windows-2019']
+        otp-version: ['24']
+        elixir-version: ['v1.12', '']
```
paulo-ferraz-oliveira, line 21: Can we move `v1.12` to `v1.14`?
Also, why is there a `''` here?

paulo-ferraz-oliveira:
I'd like to ask you for two things to ease the review process:
1. don't squash-rebase your changes (since it's harder to track what you changed) - we'll squash merge at the end
2. don't resolve the requests, but let the requester resolve them (same reason for "tracking")

paulo-ferraz-oliveira:
Are the tests failing on the Hex CDN or otherwise? I'm not sure what's behind https://cdn.jsdelivr.net/hex but that seems like the cause of all the failures.

src/install-otp.ps1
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Why the name change on the output file? Is it supposed to be known by consumers?

.github/workflows/hexpm-mirror.yml
pguyot: `''` is there to have no elixir. The logic is a little bit different when elixir is not installed: we don't cache Windows OTP and if there is no Elixir on Windows, there should be no cache.

This is why I had to modify the function that gets the Elixir version, something that will be explained in a comment as you requested [here](https://github.com/erlef/setup-beam/pull/156#discussion_r1026763634).

paulo-ferraz-oliveira:
This pull request requires a README update, where we tell consumers the option is available, and also maybe an explanation around the rationale.

src/setup-beam.js
```diff
@@ -17,21 +18,49 @@ async function main() {
   const elixirSpec = core.getInput('elixir-version', { required: false })
   const gleamSpec = core.getInput('gleam-version', { required: false })
   const rebar3Spec = core.getInput('rebar3-version', { required: false })
+  const shouldMixRebar = core.getInput('install-rebar', { required: false })
+  const shouldMixHex = core.getInput('install-hex', { required: false })
+  process.env.HEX_MIRROR = core.getInput('hexpm-mirror', { required: false })
```
paulo-ferraz-oliveira, line 23: Can we pass this as a parameter to the installer function and subsequently to the installer scripts, as is done for other stuff? Reading from the env. to have to search for what is setting a given variable is confusing and error prone.

src/setup-beam.js
```diff
@@ -17,21 +18,49 @@ async function main() {
...
+      (!elixirSpec || isVersion(elixirSpec)) &&
+      (elixirVersion || process.platform !== 'win32')
+    let hexCachedRestored = false
+    const cacheVersion = '1'
+    const hexCacheKey =
+      `hex.pm-${osVersion}-${otpVersion}-${elixirVersion}-` +
+      `r=${shouldMixRebar}-h=${shouldMixHex}-${cacheVersion}`
+    const hexCachePaths = [
+      path.join(process.env.RUNNER_TEMP, '.setup-beam', 'cache'),
+      path.join(
```
paulo-ferraz-oliveira, line 40: If we're not caching on Windows is this required?

paulo-ferraz-oliveira:
> It sounds a little bit over-engineered for this feature but why not.

I'm not sure what this means, since you answer your own question, but do you not consider it important to have CI pass all the time, even in repetition? That's the goal of your PR, right? To not have to repeat CI?

src/setup-beam.js
```diff
@@ -17,21 +18,49 @@ async function main() {
...
-    const elixirInstalled = await maybeInstallElixir(elixirSpec, otpSpec)
+    const otpVersion = await getOTPVersion(otpSpec, osVersion)
+    const elixirVersion = await getElixirVersion(elixirSpec, otpVersion)
+    // No cache for Windows OTP, rebar3 and Gleam as they're downloaded from GitHub
+    const useCache =
+      isVersion(otpSpec) &&
+      (!elixirSpec || isVersion(elixirSpec)) &&
+      (elixirVersion || process.platform !== 'win32')
+    let hexCachedRestored = false
+    const cacheVersion = '1'
```
paulo-ferraz-oliveira, line 34: What's the goal of setting this constant to only use it once, and on the line that follows?

pguyot:
> > It sounds a little bit over-engineered for this feature but why not.
> 
> I'm not sure what this means, since you answer your own question, but do you not consider it important to have CI pass all the time, even in repetition? That's the goal of your PR, right? To not have to repeat CI?

This is the goal for consumers of the action for sure. What do you suggest for this repo?

src/setup-beam.js
pguyot: The cache contains  a set of files. Shall we need to invalidate it, the version should be incremented. It seems more obvious to me to have it on a separate line.

src/setup-beam.js
pguyot: Sorry I am confused about what would not be required. Do you mean the call to the path API instead of directly using POSIX-style paths here?

This may be unrelated, but we do cache on Windows, just not Erlang/OTP.

src/setup-beam.js
pguyot: FYI the environment variable is not only used by the installer scripts, it is also used by the mix tool invoked directly from js.

pguyot:
> Are the tests failing on the Hex CDN or otherwise? I'm not sure what's behind https://cdn.jsdelivr.net/hex but (and you previously wrote "I didn't have any problem with [a](https://cdn.jsdelivr.net/hex) mirror, but I didn't test it sufficiently. My plan was to use default CDN with GH cache.") that seems like the cause of all the failures.

I had a lot of issues with repo.hex.pm.
https://cdn.jsdelivr.net/hex is a mirror listed [here](https://hex.pm/docs/mirrors).

paulo-ferraz-oliveira:
> What do you suggest for this repo?

Just thinking out loud here...

Do you know _what_ makes the download fail (e.g. a pre-condition)? (... guess not, otherwise we'd retry for a fix, right? 😄) If so, we could check that and print a proper message, instead of the cryptic one (?)

Or maybe wrap the download in a try-catch (don't know how easy it is to pinpoint a specific exception, in this case) and 1. print the stacktrace + 2. add a comment related to potential cache issue (though not sure how we'd distinguish this from the "normal" exception case). Yeah, it's tricky, I've never tested a CDN. 😁

We can leave that bit open (and keep "retrying", for now, until we have this close to concluded) and maybe some of the other maintainers has a proposal here? @starbelly?

.github/workflows/hexpm-mirror.yml
paulo-ferraz-oliveira: But I'm assuming we don't want to allow for a non-version value. In this case, you're passing in an empty string instead of `undefined`, right? Unless the code I asked to be commented is also executed in the `undefined` case. I'll wait for your comment.

paulo-ferraz-oliveira:
Keeping this for my reference: I removed the cache, re-ran the actions and the failing jobs were: 
* [Test option hexpm-mirror (ubuntu-22.04, 24, v1.12, true, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863627801#logs)
* [Test option hexpm-mirror (ubuntu-22.04, 24, v1.12, false, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863628061#logs)
* [Test option hexpm-mirror (windows-2019, 24, v1.12, false, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863629334#logs)

These are fairly similar to the ones I'd identified before (if not the same - can't know, since I deleted the cache). I'll proceed to delete the cache again and retry. It'll be odd if it fails on the same objects.

.github/workflows/hexpm-mirror.yml
pguyot: Passing an empty string here means that ${{ matrix.elixir-version }} will be replaced by an empty string, and it will be like [what happens with ubuntu.yml CI file](https://github.com/erlef/setup-beam/blob/f19acc5287cd038f617d35af560c4f3720732ad1/.github/workflows/ubuntu.yml#L144) for cases like the [first one](https://github.com/erlef/setup-beam/blob/f19acc5287cd038f617d35af560c4f3720732ad1/.github/workflows/ubuntu.yml#L24).

paulo-ferraz-oliveira: Ok, I'll wait for the comment still 👍 since I don't see why we didn't need check that before but we do now. I do have to say I'd rather have undefined or a version, than allowing for empty string, but I am aware of how strange YAML and JavaScript are in relation to empty strings, undefined and the like.

pguyot:
> Keeping this for my reference: I removed the cache, re-ran the actions and the failing jobs were:
> 
> * [Test option hexpm-mirror (ubuntu-22.04, 24, v1.12, true, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863627801#logs)
> * [Test option hexpm-mirror (ubuntu-22.04, 24, v1.12, false, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863628061#logs)
> * [Test option hexpm-mirror (windows-2019, 24, v1.12, false, true, https://cdn.jsdelivr.net/hex)](https://github.com/erlef/setup-beam/actions/runs/3499133871/jobs/5863629334#logs)
> 
> These are fairly similar to the ones I'd identified before (if not the same - can't know, since I deleted the cache). I'll proceed to delete the cache again and retry. It'll be odd if it fails on the same objects.

In my experience it really depends on the day / time of day.

The network link between GitHub runners and repo.hex.pm has been more stable recently but it was such a nightmare that I did file this PR in the first place.

src/setup-beam.js
paulo-ferraz-oliveira: I see. Can we then not do a `HEX_MIRROR=... mix` from that call, in this case? I'd really prefer to not have to use env. variables to pass values around.

pguyot:
> Or maybe wrap the download in a try-catch (don't know how easy it is to pinpoint a specific exception, in this case) and 1. print the stacktrace + 2. add a comment related to potential cache issue (though not sure how we'd distinguish this from the "normal" exception case). Yeah, it's tricky, I've never tested a CDN. 😁

Alternatively, we could change yet again the logic and work with several mirrors (shall they be settable as an option?) and try them in order.

src/setup-beam.js
paulo-ferraz-oliveira: To invalidate the caches shouldn't you just delete them, from (e.g. in our case) https://github.com/erlef/setup-beam/actions/caches?

src/setup-beam.js
paulo-ferraz-oliveira: Ah, I must've been confused by a previous comment. I though we weren't caching on Windows at all, but if it's just OTP, it's Ok to keep this. Resolving...

src/setup-beam.js
pguyot: Consumers of the action wouldn't know when to invalidate the cache.

The other option would be to get a hash from the source code of the file as is often done with the cache action. I didn't find anything simple to do that without adding more dependencies though.

paulo-ferraz-oliveira:
> Alternatively, we could change yet again the logic and work with several mirrors (shall they be settable as an option?) and try them in order.

If that's not too big of a change, it seems reasonable. After you try at least two (I imagine that might be a documented normal case) we could then write a proper warning (?)

Instead of `hexpm-mirror` (string) we could have `hexpm-mirrors` (array of strings).

.github/workflows/hexpm-mirror.yml
pguyot: See also https://github.com/actions/toolkit/blob/819157bf872a49cfcc085190da73894e7091c83c/packages/core/src/core.ts#L120

README.md
```diff
@@ -189,6 +189,13 @@ jobs:
 
 **Note**: the `otp-version: false` input is only applicable when installing Gleam.
 
+## Hex.pm mirrors and caching
+
+It is possible to define the hex.pm mirrors to use including their order with
+option `hexpm-mirrors`. By default, the action will fetch from `repo.hex.pm`
+first and then from `cdn.jsdelivr.net/hex`. Permanent files from these mirrors,
```
ericmj, line 196: This should not be a default, cdn.jsdelivr.net/hex is not an official mirror and we cannot give an any guarantees for it. Unlike when download Hex packages there is no public key verification when downloading builds. It is also known to sometimes lag behind by a few days on certain paths.

I think retries are a great idea but should be done against the official repo and if users want to use a different repo or multiple with round robin then that should be opt in.

README.md
```diff
@@ -189,6 +189,13 @@ jobs:
...
+such as OTP binaries for a given version, are cached automatically.
```
ericmj, line 197: Is there a way to bust the cache?

src/setup-beam.js
```diff
@@ -17,20 +20,55 @@ async function main() {
...
   if (otpSpec !== 'false') {
-    await installOTP(otpSpec, osVersion)
-    const elixirInstalled = await maybeInstallElixir(elixirSpec, otpSpec)
+    const otpVersion = await getOTPVersion(otpSpec, osVersion, hexMirrors)
+    const elixirVersion = await getElixirVersion(
+      elixirSpec,
+      otpVersion,
+      hexMirrors,
+    )
+    // No cache for Windows OTP, rebar3 and Gleam as they're downloaded from GitHub
```
ericmj, line 36: Why do we not cache things from GitHub?

pguyot: As the cache is on GitHub, what's the point of caching GitHub?

README.md
pguyot: The cache keys depends on the MD5 sum of `setup-beam.js`.

README.md
pguyot: Thank you for your review.

I am confused about these two points:

> Unlike when download Hex packages there is no public key verification when downloading builds.

Are you referring to the mechanism in [Mix.Local](https://github.com/elixir-lang/elixir/blob/d9c7e80302c220a2b603ac16931dee6056a6c863/lib/mix/lib/mix/local.ex#L144)?

I am confused how this wouldn't apply to downloads from `cdn.jsdelivr.net/hex`. I had a quick look at the code, which obviously you contributed to, and it seems to me that mix also verifies the signature with the mirror by using signature csv files for [hex](http://cdn.jsdelivr.net/hex/installs/hex-1.x.csv) and for [rebar](https://cdn.jsdelivr.net/hex/installs/rebar3-1.x.csv).

> It is also known to sometimes lag behind by a few days on certain paths.

In the current implementation, for Linux OTP and Elixir builds, the action first tries to fetch the list of builds and it does so in the order of the mirrors, so starting with the official mirror. And then it proceeds with downloading the binary, proceeding likewise in the order of mirrors. The only case where it would pick an older version (compared to the official mirror) would be if the official mirror didn't respond properly.

For consumers of the action, do we want by default to fail a build because `repo.hex.pm` is not available or fails (which it did a lot recently) without trying next with a mirror that might save the build?

The whole point of this PR is to make builds that depend on this action more robust. It seems to me that defaulting to a maybe lagging mirror is a good solution. If consumers want a particular version, they can mention it in the spec and if this version is not available in the mirror, the build will obviously fail.

README.md
ericmj: Is there a way for the user to bust the cache? If there is I think it would be useful to add to the documentation.

src/setup-beam.js
ericmj: The machines that host github action runners are not on the same network as github artifact hosting. The artifact hosting uses [Fastly CDN](https://www.fastly.com/) which we also use for repo.hex.pm. If there is network connectivity issues when trying to reach repo.hex.pm then github artifacts will have the same issues unless they have an extra layer of caching I don't know about.

README.md
ericmj: > Are you referring to the mechanism in [Mix.Local](https://github.com/elixir-lang/elixir/blob/d9c7e80302c220a2b603ac16931dee6056a6c863/lib/mix/lib/mix/local.ex#L144)?

`mix local.hex` and `mix local.rebar` will be verified but the Elixir and Erlang installations are not installed through Mix and will not be verified.

> For consumers of the action, do we want by default to fail a build because repo.hex.pm is not available or fails (which it did a lot recently) without trying next with a mirror that might save the build?

I think we should retry our fetching from repo.hex.pm X number of times. We don't know if trying another mirror will be more successful than retrying the original mirror. I don't think we should default to falling back to a mirror we have no control over, can provide no guarantees for, and we cannot verify the integrity of the files we dowload.

Adding support for additional mirrors is great but if we pick an unofficial mirror to use as default I would like to see integrity checks for the files we downloads and stats that show that retrying to it will have a higher success rate.

.github/workflows/hexpm-mirrors.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          otp-version: '24.3.4.6'
+          elixir-version: 'v1.13.4'
+          install-rebar: true
+          install-hex: true
+          hexpm-mirrors: https://mirror.invalid
+      - name: Report unexpected success
+        if: ${{ steps.failing-setup-beam.outcome == 'success' }}
+        run: |
+          echo "Expected erlef-setup beam task to fail"
```
paulo-ferraz-oliveira, line 57:
```suggestion
          echo "Action is expected to fail"
```

README.md
paulo-ferraz-oliveira: I agree with @ericmj.

The **default**, if the option is not used, should be whatever behaviour we had before (except for the obvious cache).

The fact is nobody else (up until now, at least) had similar issues (that were reported), so the option itself should be opt-in.

On the other hand, if you do decide to opt-in maybe 3 tries, with exponential fallback would be Ok in a round-robin fashion. What do you think, @ericmj?

e.g. 

```
hexpm-mirrors:
  - A
  - B
  - C
```

would either do A, A, A, B, B, B, C, C, C, or A, B, C, A, B, C, ... with exponential backoff every time a previous mirror is retried.

Also, @pguyot, it would be nice to add a "how to use it" to the README. I don't expect consumers of the action to have to go to the code for that. There's an example section at the bottom that might be appropriate for an example.

README.md
paulo-ferraz-oliveira: It'll be on the consumer, right? i.e. this cache will appear as a normal cache the consumer can delete from their list of caches... If that's so we can maybe point to either something like .../actions/caches, or to GitHub Actions caches' documentation: https://docs.github.com/en/rest/actions/cache#list-github-actions-caches-for-a-repository.

src/setup-beam.js
```diff
@@ -173,13 +198,25 @@ async function getOTPVersion(otpSpec0, osVersion) {
   return otpVersions.get(otpVersion) // from the reference, for download
 }
 
-async function getElixirVersion(exSpec0, otpVersion0) {
+/**
+ * Return the elixir version from a specification and an OTP version.
```
paulo-ferraz-oliveira, line 202:
```suggestion
 * Return the Elixir version from a specification and an OTP version.
```

src/setup-beam.js
```diff
@@ -173,13 +198,25 @@ async function getOTPVersion(otpSpec0, osVersion) {
...
+ *
+ * Process the OTP version and the Elixir specification to find the version
+ * in the list of available elixir versions.
```
paulo-ferraz-oliveira, line 205:
```suggestion
 * in the list of available Elixir versions.
```

src/setup-beam.js
```diff
@@ -173,13 +198,25 @@ async function getOTPVersion(otpSpec0, osVersion) {
...
+ *
+ * @param exSpec0 Elixir specification. Can be the empty string.
```
paulo-ferraz-oliveira, line 207: What does the empty string serve apart from the test that was included to have it there? Can consumers now simply not have a version? I don't understand the "why" (which is what I'd expect I would).

action.yml
```diff
@@ -33,9 +33,11 @@ inputs:
   disable_problem_matchers:
     description: whether to have the problem matchers present in the results (or not)
     default: false
-  hexpm-mirror:
+  hexpm-mirrors:
     description: mirror for hex.pm
```
paulo-ferraz-oliveira, line 37:
```suggestion
    description: mirrors for hex.pm
```

README.md
ericmj: I think `A, A, A, B, B, B, C, C, C` should be the behavior.

.github/workflows/hexpm-mirror.yml
paulo-ferraz-oliveira: Ok, the good reason to justify this change was that the code was previously protected at https://github.com/erlef/setup-beam/pull/156/files#diff-6406ae11684199b17d2cdca1efe2cfa29fb4587c3b142f1d33ad5f81a5acf7adL60, and now `getElixirVersion` was moved out of that function.

src/setup-beam.js
paulo-ferraz-oliveira: It seems you did, after all, using just `crypto`.

README.md
paulo-ferraz-oliveira: This is apparently now done via `crypto`, as per https://github.com/erlef/setup-beam/pull/156/files#diff-6406ae11684199b17d2cdca1efe2cfa29fb4587c3b142f1d33ad5f81a5acf7adR42 and subsequent lines of code...

src/setup-beam.js
paulo-ferraz-oliveira: The rationale behind this is now explained in another comment. Long story short... a change to the code exposed this, whereas before it was hidden behind another `if`.

paulo-ferraz-oliveira:
@pguyot, I think this is getting close to ready.

We're just missing some small things:

- [ ] [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1026960101)
- [ ] [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028523824)
- [ ] [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028515853)
- [ ] [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530097)
- [ ] [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530246)
- [ ] [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028532269)
- [ ] [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028533044)
- [ ] finally, there's now merge conflicts

src/setup-beam.js
pguyot: > @pguyot, I think this is getting close to ready.
> 
> We're just missing some small things:
> 
> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1026960101)
This was fixed  in 634fc94ae5b9d0c83f7ff2928dfbec41ddddc9ed.

> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028523824)

Sorry, this PR is not about retries. We shouldn't retry mirrors.
In fact, I even removed the caching feature as it doesn't make sense for mirrors as well.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028515853)
Commit created.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530097)
Because cache feature has been removed, it is no longer necessary to modify the semantics of this function and therefore I removed the documentation I wrote.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530246)
Likewise.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028532269)
Outdated.

> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028533044)
This is how options work with actions. The API provides a multiline parsing. I believe lists are not supported.

> * [ ]  finally, there's now merge conflicts
My approach is to rebase and force push, but this is not what you asked for. Please advise.

pguyot:
> @pguyot, I think this is getting close to ready.
> 
> We're just missing some small things:
> 
> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1026960101)
This was fixed  in 634fc94ae5b9d0c83f7ff2928dfbec41ddddc9ed.

> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028523824)

Sorry, this PR is not about retries. We shouldn't retry mirrors.
In fact, I even removed the caching feature as it doesn't make sense for mirrors as well.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028515853)
Commit created.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530097)
Because cache feature has been removed, it is no longer necessary to modify the semantics of this function and therefore I removed the documentation I wrote.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028530246)
Likewise.

> * [ ]  [update suggestion](https://github.com/erlef/setup-beam/pull/156#discussion_r1028532269)
Outdated.

> * [ ]  [comment](https://github.com/erlef/setup-beam/pull/156#discussion_r1028533044)
This is how options work with actions. The API provides a multiline parsing. I believe lists are not supported.

> * [ ]  finally, there's now merge conflicts
My approach is to rebase and force push, but this is not what you asked for. Please advise.

README.md
```diff
@@ -189,12 +189,47 @@ jobs:
 
 **Note**: the `otp-version: false` input is only applicable when installing Gleam.
 
-## Hex.pm mirrors and caching
+## Hex.pm mirrors
```
paulo-ferraz-oliveira, line 192: This section should be moved before "License".

README.md
```diff
@@ -189,6 +189,48 @@ jobs:
 
 **Note**: the `otp-version: false` input is only applicable when installing Gleam.
 
+## Hex.pm mirrors
+
+It is possible to define the hex.pm mirror(s) to use including their order with
```
paulo-ferraz-oliveira, line 194:
```suggestion
It is possible to define the hex.pm mirror(s) to use, in their declared order, with
```

action.yml
```diff
@@ -33,6 +33,9 @@ inputs:
   disable_problem_matchers:
     description: whether to have the problem matchers present in the results (or not)
     default: false
+  hexpm-mirrors:
+    description: mirror(s) for hex.pm, one per line
+    default: https://repo.hex.pm
```
paulo-ferraz-oliveira, line 38:
```suggestion
    default: |
      https://repo.hex.pm
```

paulo-ferraz-oliveira:
> My approach is to rebase and force push, but this is not what you asked for. Please advise.

I believe we will be able to rebase+force-push once everything's dealt with. I added it to my list of TO-DO because it's something that's missing.

FWIW, this pull request is changing context and scope so often is difficult to keep track of what's requested. I'd advise to have a discussion with the maintainers if you intend to change something else (that's not requested by us) in order to ease the review process.

package.json
```diff
@@ -19,6 +19,7 @@
     }
   },
   "dependencies": {
+    "@actions/cache": "3.0.6",
```
paulo-ferraz-oliveira, line 22: Are we still using this? If not, it should be reversed.

paulo-ferraz-oliveira:
I'd prefer repo.hex.pm to not only be the default, but the fallback in any case - except if it's already part of the chosen mirrors, especially because it's the official place we get the builds from, at the moment. This way we don't need to fail for an empty mirror list, and by renaming the option as `alternative-...` we make it clear what we're offering.

src/installer.js
```diff
@@ -6,11 +7,28 @@ const path = require('path')
...
+ * @param {string[]} hexMirrors
  */
-async function installOTP(osVersion, otpVersion) {
+async function installOTP(osVersion, otpVersion, hexMirrors) {
   const OS = process.platform
   if (OS === 'linux') {
-    await exec(path.join(__dirname, 'install-otp.sh'), [osVersion, otpVersion])
+    if (hexMirrors.length === 0) {
+      throw new Error(
+        `Could not install OTP ${osVersion} ${otpVersion} from any hex mirror`,
```
paulo-ferraz-oliveira, line 17:
```suggestion
        `Could not install OTP ${otpVersion} from any hex.pm mirror`,
```

src/installer.js
```diff
@@ -21,15 +39,35 @@ async function installOTP(osVersion, otpVersion) {
  * Install Elixir.
  *
  * @param {string} elixirVersion
+ * @param {string[]} hexMirrors
  */
-async function installElixir(elixirVersion) {
+async function installElixir(elixirVersion, hexMirrors) {
+  if (hexMirrors.length === 0) {
+    throw new Error(
+      `Could not install Elixir ${elixirVersion} from any hex mirror`,
```
paulo-ferraz-oliveira, line 47:
```suggestion
      `Could not install Elixir ${elixirVersion} from any hex.pm mirror`,
```

src/installer.js
```diff
@@ -21,15 +39,35 @@ async function installOTP(osVersion, otpVersion) {
...
+    }
+    if (OS === 'win32') {
+      const script = path.join(__dirname, 'install-elixir.ps1')
+      await exec(
+        `pwsh.exe ${script} -VSN:${elixirVersion} -HEX_MIRROR:${hexMirror}`,
+      )
+      return
+    }
+  } catch (err) {
+    core.info(`install-elixir failed for mirror ${hexMirror}`)
```
paulo-ferraz-oliveira, line 68: We should output the name of the script (by a variable?) so that's its the actual one that failed.

paulo-ferraz-oliveira:
> My approach is to rebase and force push, but this is not what you asked for. Please advise.

We'll do the rebase, resolve conflict, and force push at the end.

src/setup-beam.js
```diff
@@ -17,20 +17,27 @@ async function main() {
   const elixirSpec = core.getInput('elixir-version', { required: false })
   const gleamSpec = core.getInput('gleam-version', { required: false })
   const rebar3Spec = core.getInput('rebar3-version', { required: false })
+  const hexMirrors = core.getMultilineInput('hexpm-mirrors', {
```
paulo-ferraz-oliveira, line 20: As proposed, in a different comment, we should:
1. have hexMirrors be repo.hex.pm if the input is present but empty
2. have hexMirrors always contain repo.hex.pm (if not already present) at the end, as the last fallback

src/setup-beam.js
```diff
@@ -79,12 +90,28 @@ async function maybeInstallElixir(elixirSpec, otpVersion) {
   return installed
 }
 
-async function mix(shouldMix, what) {
+async function mixWithMirrors(cmd, args, hexMirrors) {
+  if (hexMirrors.length === 0) {
+    throw new Error('mix failed with every mirror')
+  }
+  const [hexMirror, ...hexMirrorsT] = hexMirrors
+  process.env.HEX_MIRROR = hexMirror
```
paulo-ferraz-oliveira, line 98: Is setting this env. variable still required?

src/setup-beam.js
```diff
@@ -79,12 +90,28 @@ async function maybeInstallElixir(elixirSpec, otpVersion) {
...
+  try {
+    return await exec(cmd, args)
+  } catch (err) {
+    core.info(
+      `Mix failed with mirror ${process.env.HEX_MIRROR} with message ${err.message})`,
```
paulo-ferraz-oliveira, line 103:
```suggestion
      `mix failed with mirror ${process.env.HEX_MIRROR} with message ${err.message})`,
```

README.md
paulo-ferraz-oliveira, line 194:
```suggestion
It is possible to use alternative hex.pm mirror(s), in their declared order, with
```

README.md
```diff
@@ -189,6 +189,48 @@ jobs:
 
 **Note**: the `otp-version: false` input is only applicable when installing Gleam.
 
+## Hex.pm mirrors
```
paulo-ferraz-oliveira, line 192:
```suggestion
## Alternative Hex.pm mirrors
```

README.md
```diff
@@ -189,6 +189,48 @@ jobs:
...
+
+It is possible to define the hex.pm mirror(s) to use including their order with
+option `hexpm-mirrors`. By default, the action will fetch from `repo.hex.pm`.
+To define several mirrors, simply put each in a given line.
```
paulo-ferraz-oliveira, line 196:
```suggestion
To use other alternative mirrors, add one per line, as shown below.
```

README.md
```diff
@@ -189,6 +189,48 @@ jobs:
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: '25'
+          hexpm-mirrors: https://cdn.jsdelivr.net/hex
```
paulo-ferraz-oliveira, line 212:
```suggestion
          # Use `cdn.jsdelivr.net/hex` before falling back to `repo.hex.pm`
          hexpm-mirrors: https://cdn.jsdelivr.net/hex
```

action.yml
paulo-ferraz-oliveira, line 38:
```suggestion
    default: |
      https://repo.hex.pm
```

action.yml
```diff
@@ -33,6 +33,9 @@ inputs:
   disable_problem_matchers:
     description: whether to have the problem matchers present in the results (or not)
     default: false
+  hexpm-mirrors:
+    description: mirror(s) for hex.pm, one per line
```
paulo-ferraz-oliveira, line 37:
```suggestion
    description: alternative hex.pm mirror(s)
```

README.md
```diff
@@ -189,6 +189,48 @@ jobs:
...
 **Note**: the `otp-version: false` input is only applicable when installing Gleam.
 
+## Hex.pm mirrors
+
+It is possible to define the hex.pm mirror(s) to use including their order with
+option `hexpm-mirrors`. By default, the action will fetch from `repo.hex.pm`.
+To define several mirrors, simply put each in a given line.
+
+To use `cdn.jsdelivr.net/hex` instead of `repo.hex.pm`:
+
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: This is removed as we'll be adding a comment in the YML, below.

.github/workflows/hexpm-mirrors.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test_failing_first_mirror:
```
paulo-ferraz-oliveira, line 13:
```suggestion
  test-failing-first-mirror:
```

.github/workflows/hexpm-mirrors.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: ./
+        with:
+          otp-version: ${{matrix.otp-version}}
+          elixir-version: ${{matrix.elixir-version}}
+          install-rebar: ${{matrix.install-rebar}}
+          install-hex: ${{matrix.install-hex}}
+          hexpm-mirrors: |
+            https://mirror.invalid
+            https://cdn.jsdelivr.net/hex
+            https://repo.hex.pm
```
paulo-ferraz-oliveira, line 37: Not required, if it's the last one, but having it for clarity should not change the final list (which would only append if not present).

.github/workflows/hexpm-mirrors.yml
```diff
@@ -0,0 +1,58 @@
...
+
+  test_invalid_mirror:
```
paulo-ferraz-oliveira, line 39:
```suggestion
  test-invalid-mirror:
```

.github/workflows/hexpm-mirrors.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Use erlef/setup-beam
+        id: failing-setup-beam
+        uses: ./
+        continue-on-error: true
+        with:
+          otp-version: '24.3.4.6'
+          elixir-version: 'v1.13.4'
+          install-rebar: true
+          install-hex: true
+          hexpm-mirrors: https://mirror.invalid
```
paulo-ferraz-oliveira, line 53:
```suggestion
          alternative-hexpm-mirrors: |
            https://mirror.invalid
```

src/setup-beam.js
```diff
@@ -449,6 +478,19 @@ async function get(url0, pageIdxs) {
   return ret
 }
 
+async function getWithMirrors(resourcePath, hexMirrors) {
+  if (hexMirrors.length === 0) {
+    throw new Error(`Could not fetch ${resourcePath} from any hex mirror`)
```
paulo-ferraz-oliveira, line 483:
```suggestion
    throw new Error(`Could not fetch ${resourcePath} from any hex.pm mirror`)
```

src/setup-beam.js
pguyot: This doesn't follow mix's `HEX_MIRROR` usage.
This PR is also to allow users to entirely opt out of `repo.hex.pm`, for example on hosted runners. Forcing users to always use `repo.hex.pm` doesn't seem right to me.

paulo-ferraz-oliveira: I'll update my request then: the action shall stop with error if `alternative-hexpm-mirrors` is empty (if this is even possible in itself).

src/setup-beam.js
paulo-ferraz-oliveira: Ok, I'll suggest a change to add a comment, then... thank you.

paulo-ferraz-oliveira, line 98:
```suggestion
  // As per https://hex.pm/docs/mirrors
  process.env.HEX_MIRROR = hexMirror
```

src/setup-beam.js
paulo-ferraz-oliveira: Please don't resolve comments, to ease the review process, as previously requested.

src/setup-beam.js
paulo-ferraz-oliveira: Will be resolved via a comment.

README.md
paulo-ferraz-oliveira: We dropping retries, and changing the implemented mechanism, so this conversation is no longer relevant.

src/setup-beam.js
pguyot: GitHub automatically resolves comments when the suggestion is committed.

paulo-ferraz-oliveira:
@pguyot, do you intend to continue working on this?

pguyot:
> @pguyot, do you intend to continue working on this?

I can rebase and fix conflicts to clarify the intent of this PR. It really is about mirroring (pun intended) the behavior of mix's HEX_MIRROR's environment variable instead of default repo.hex.pm.

PR #159, opened 2022-11-19
Ease consumption by means of .tool-versions

### LINKED ISSUE #137: Add support for .tool-versions files
Thank you for this excellent Github action!

Most of the time, I want the action to run the same environment that I use locally, which I specify with the asdf version manager that many others use for Elixir and Erlang. It would be convenient to just set the `version-file` key or something similar to `.tool-versions` and then have it automatically resolve the values from there. Setting this value would also necessarily set `version-type` to `strict` as well.

This can be done manually, probably, by wiring up another external action that does this reading, then passing along the values to this action. But Github actions are finicky enough that I think there is a benefit to having this functionality baked into this action.

This is a feature that is similar to what the `setup-node` action supports. 

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

paulo-ferraz-oliveira:
This allows using a `.tool-versions` file as per `asdf` constraints.

The `README.md` changes should make the content of the pull request explicit.

---

**Edit**

This is working as expected as can be seen for example repo ([link to `.tool-versions`](https://github.com/paulo-ferraz-oliveira/setup-beam-example/blob/main/.tool-versions)):

- here: [success / versions installed as per `.tool-versions`](https://github.com/paulo-ferraz-oliveira/setup-beam-example/actions/runs/3501999438/jobs/5866104338#step:3:20), 
- here: [error / `version-type =/= strict`](https://github.com/paulo-ferraz-oliveira/setup-beam-example/actions/runs/3501999438/jobs/5866104338#step:4:13)
- and here: [error / colliding inputs](https://github.com/paulo-ferraz-oliveira/setup-beam-example/actions/runs/3501999438/jobs/5866104338#step:5:20)

---

Closes #137.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -40,6 +40,11 @@ jobs:
           node-version: '14'
       - run: npm ci
       - run: npm test
+      - name: .tool-versions test
+        id: setup-beam
+        uses: ./
+        with:
+          version-file: '__tests__/.tool-versions'
 
   unit_tests_windows:
     name: Unit tests (Windows)
@@ -51,3 +56,8 @@ jobs:
           node-version: '14'
       - run: npm install --production
       - run: npm test
+      - name: .tool-versions test
+        id: setup-beam
+        uses: ./
+        with:
+          version-file: '__tests__/.tool-versions'
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -57,9 +57,3 @@ jobs:
           node-version: '14'
       - run: npm install --production
       - run: npm test
-      - name: .tool-versions test
-        id: setup-beam
-        uses: ./
-        with:
-          version-file: __tests__/.tool-versions
-          version-type: strict
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -44,7 +44,8 @@ jobs:
         id: setup-beam
         uses: ./
         with:
-          version-file: '__tests__/.tool-versions'
+          version-file: __tests__/.tool-versions
+          version-type: strict
 
   unit_tests_windows:
     name: Unit tests (Windows)
@@ -60,4 +61,5 @@ jobs:
         id: setup-beam
         uses: ./
         with:
-          version-file: '__tests__/.tool-versions'
+          version-file: __tests__/.tool-versions
+          version-type: strict
--- a/README.md
+++ b/README.md
@@ -16,9 +16,11 @@ workflow by:
 - optionally, installing [Gleam](https://gleam.run/)
 - optionally, installing [`rebar3`](https://www.rebar3.org/)
 - optionally, installing [`hex`](https://hex.pm/)
+- optionally, opting for strict or loose version matching
 - optionally, having
   [problem matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) show
   warnings and errors on pull requests
+- optionally, using a version file (as explained in "Version file", below), to identify versions
 
 **Note**: currently, this action only supports Actions' `ubuntu-` and `windows-` runtimes.
 
@@ -85,6 +87,31 @@ jobs:
       - uses: erlef/setup-beam@v1
         ...
 ```
+
+### Version file
+
+A version file is specified via input `version-file` (e.g.`.tool-versions`). This
+allows not having to use YML input for versions, though the action does check (and
+will exit with error) if both inputs are set.
+
+**Note**: if you're using a version file, option `version-type` is checked to be `strict`,
+and will make the action exit with error otherwise.
+
+The following version file formats are supported:
+
+- `.tool-versions`, as specified by [asdf: Configuration](https://asdf-vm.com/manage/configuration.html)
+
+Supported version elements are the same as the ones defined for the YML portion of the action,
+with the following correspondence.
+
+#### `.tool-versions` format
+
+| YML              | `.tool-versions` |
+|-                 |-
+| `otp-version`    | `erlang`
+| `elixir-version` | `elixir`
+| `gleam-version`  | `gleam`
+| `rebar3-version` | `rebar`
 
 ### Example (Erlang/OTP + Elixir, on Ubuntu)
 
--- a/action.yml
+++ b/action.yml
@@ -31,8 +31,12 @@ inputs:
       to guess versions based on semver rules
     default: loose
   disable_problem_matchers:
-    description: whether to have the problem matchers present in the results (or not)
+    description: whether to have the problem matchers present in the results
+      (or not)
     default: false
+  version-file:
+    description: a versions file (e.g. as used by `asdf`), which defines inputs
+    default: ''
 outputs:
   elixir-version:
     description: Exact version of Elixir that was installed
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -16,9 +16,11 @@ workflow by:
 - optionally, installing [Gleam](https://gleam.run/)
 - optionally, installing [`rebar3`](https://www.rebar3.org/)
 - optionally, installing [`hex`](https://hex.pm/)
+- optionally, opting for strict or loose version matching
```
paulo-ferraz-oliveira, line 19: Not specific to this pull request, but was missing.

src/setup-beam.js
```diff
@@ -460,10 +471,62 @@ function isVersion(v) {
   return /^v?\d+/.test(v)
 }
 
+function getInput(inputName, required, alternativeName, alternativeValue) {
+  // We can't have both input and alternativeValue set
+  let input = core.getInput(inputName, {
+    required: alternativeValue ? false : required,
+  })
```
paulo-ferraz-oliveira, line 478:
```suggestion
  let input = core.getInput(inputName, {
    required: alternativeValue ? false : required,
  })
  // We can't have both input and alternativeValue set
```

src/setup-beam.js
```diff
@@ -460,10 +471,62 @@ function isVersion(v) {
...
+    )
+  }
+  console.log(`##[group]Parsing version file at ${versionFilePath0}`)
+  const appVersions = new Map()
+  const versions = fs.readFileSync(versionFilePath, 'utf8')
+  versions.split('\n').forEach((line) => {
+    const appVersion = line.match(/^([^ ]+)[ ]+([^ #]+)/)
+    if (appVersion) {
+      const app = appVersion[1]
+      if (['erlang', 'elixir', 'gleam', 'rebar3'].includes(app)) {
```
paulo-ferraz-oliveira, line 508:
```suggestion
      if (['erlang', 'elixir', 'gleam', 'rebar'].includes(app)) {
```

action.yml
```diff
@@ -11,7 +11,6 @@ inputs:
   otp-version:
     description: Version range or exact version of Erlang/OTP to use,
       or false when installing only Gleam without OTP
-    required: true
```
paulo-ferraz-oliveira, line 14: This is check by code, in any case

paulo-ferraz-oliveira:
Moving to Review-ready.

@kevinlang

.github/workflows/action.yml
```diff
@@ -40,6 +40,12 @@ jobs:
           node-version: '14'
       - run: npm ci
       - run: npm test
+      - name: .tool-versions test
+        id: setup-beam
+        uses: ./
+        with:
+          version-file: __tests__/.tool-versions
+          version-type: strict
```
wojtekmach, line 48: I understand the rationale but I think it would be more convenient if we didn't have to specify version-type if version-file is set. (we'd fail if it is set to something else than strict though.) Not a strong opinion.

wojtekmach (commented):
I'm personally not a fan of projects that commit `.tool-versions` as they would overwrite my global Elixir version choice. We wouldn't be able to easily support a valid asdf config like: `elixir: 1.14.2-otp-25`. I'm personally -1 on adding this but I don't have a strong opinion. For people who are checking in the versions file, it is not easy to parse and fill in the setup-beam version values in the workflow, so this change is definitely very convenient.

.github/workflows/action.yml
starbelly: Are there actually cases where people put their `.tool-versions` deeper in their project? I think convention over configuration would indeed be better here, and then you provide a boolean. `versions-from-file: true` or something like that.

starbelly:
> I'm personally not a fan of projects that commit `.tool-versions` as they would overwrite my global Elixir version choice. We wouldn't be able to easily support a valid asdf config like: `elixir: 1.14.2-otp-25`. I'm personally -1 on adding this but I don't have a strong opinion. For people who are checking in the versions file, it is not easy to parse and fill in the setup-beam version values in the workflow, so this change is definitely very convenient.

I'm with you here, I see this as mostly useful for private applications myself. Not too many libs out in the wild using `.tool-versions` AFAIK.

.github/workflows/action.yml
paulo-ferraz-oliveira: `version-type` is `loose` by default, so not specifying it is the same as choosing non-strict. I guess validating, in that case, unless it is `strict`, would always fail (?).

There is, however, a [note on the README for this](https://github.com/erlef/setup-beam/pull/159/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R97), so it'll fail at most once on your first run, where you include the new option. This seems safe.

@starbelly, as for a new option alongside a new option, I don't understand the rationale.

> Are there actually cases where people put their .tool-versions deeper in their project?

It wouldn't matter, because the action reads a path, so you can put it wherever you want. The fact I named it like that, taking into account the issue that gave origin to this pull request, is that we might support more conventions in the future. If we think we won't, we should just change the option name to `parse-tool-versions` and use a default as you propose.

paulo-ferraz-oliveira:
From what you wrote, I have a feeling it doesn't hurt, but none of you are inclined to accept it, is that it, @starbelly, @wojtekmach? I guess if that's the case we need a better discussion workflow. 😄 There was [an issue around this](https://github.com/erlef/setup-beam/issues/137), and the idea wasn't reject by anyone. I later wrote I'd start to implement it, and [I was apparently encouraged](https://github.com/erlef/setup-beam/issues/137#issuecomment-1287939479), but I might have over-read it, or under-reading what you wrote here.

I don't mind closing this pull request and moving forward, but I would like more concrete opinions on whether we keep it or drop it. 👍

kevinlang:
I think having `.tools-versions` checked into the project is mainly beneficial for private applications, as @starbelly noted. I don't think disliking the default ASDF behavior or respecting `.tools-versions` is a sufficient reason for excluding this feature. If one does not like that ASDF behavior, the solution is to configure their ASDF to not respect repository-local `.tools-versions` file. This can be done as outlined in [this comment](https://github.com/asdf-vm/asdf/issues/870)

wojtekmach:
Sorry I wasn't clear. I personally wouldn't add it for the reason mentioned. But I'm not opposed to adding it. Does that make sense?

wojtekmach (approved):

paulo-ferraz-oliveira:
> Sorry I wasn't clear. I personally wouldn't add it for the reason mentioned. But I'm not opposed to adding it. Does that make sense?

It does, thanks :)

paulo-ferraz-oliveira:
And I know there might be this tendency to say "but it's optional", etc., but options have to be maintained too, so I do understand the rationale behind your rationale. (and also that we might run into issues regarding proper parsing of the tools file).

.github/workflows/action.yml
starbelly: @paulo-ferraz-oliveira Eh, it's not really a big deal IMO. I think the way you have it is more flexible, what  I just suggested would result in almost the same amount of typing 😁

starbelly (approved):

starbelly:
I just wanted to say great job on this @paulo-ferraz-oliveira , there's a lot of teams that will get a lot of mileage out of it! ❤️

paulo-ferraz-oliveira:
I'm going to check deeper how the file is parsed by `asdf` (I only did it by trying, for this implementation, initially) and replicate it. The magic happens here: https://github.com/asdf-vm/asdf/blob/master/lib/utils.bash#L652.

PR #168, opened 2022-12-20
Try to prevent 403 when accessing the GitHub API

### LINKED ISSUE #167: Frequent failures when HTTP-calling GitHub
Hello! This month I've started to get CI failures very frequently when the action makes requests to the github API with `erlef/setup-beam@v1.15.0`

```
Error: Got 403 from https://api.github.com/repos/erlang/rebar3/releases?per_page=100&page=2. Exiting with error
```

https://github.com/gleam-lang/gleam/actions/runs/3725950844/jobs/6319069926

https://github.com/gleam-lang/gleam/actions/runs/3726000010/jobs/6319157464

https://github.com/gleam-lang/gleam/actions/runs/3726020768/jobs/6319192244

It is a transient error, if the workflow is retried it succeeds. I'm unsure what the cause is.

paulo-ferraz-oliveira:
Closes #167.

**Edit**: I wanted to open this on my fork, but got duped by GitHub once more and opened it here. It's a draft, for the time being.

- [x] Make sure we're not doing `JSON.parse` if the response is not JSON (or otherwise catch that exception to make it more usable for consumers)


paulo-ferraz-oliveira:
Moving to "Ready to review", to open it to reviewers, but don't merge just yet. I wanna make sure that running CI repeatedly does not surface any issues. (and I'm still missing a self-review)

.github/workflows/action.yml
```diff
@@ -28,7 +31,7 @@ jobs:
       - run: npm run yamllint
       - run: npm run jslint
       - name: Check if build left artifacts
-        run: git diff --exit-code
+        run: git diff --exit-code --ignore-space-at-eol
```
paulo-ferraz-oliveira, line 34: I get strange behaviour without this, in that the same number of additions and subtractions is present in the diff, but it exits with a code other than 0. If somebody wants me to dig deeper, I could, but I'm not sure how relevant this is in our context (the goal here is making sure that whatever is pushed is what is present locally when running `npm run build-dist`).

src/setup-beam.js
```diff
@@ -413,38 +413,37 @@ function getRunnerOSVersion() {
 }
 
 async function get(url0, pageIdxs) {
-  function getPage(pageIdx) {
+  function getPage(pageIdx, retries0) {
+    let retries = retries0
     return new Promise((resolve, reject) => {
+      if (retries === undefined) {
+        retries = 5
```
paulo-ferraz-oliveira, line 420: We retry up to 5 times, on failure.

src/setup-beam.js
```diff
@@ -413,38 +413,37 @@ function getRunnerOSVersion() {
...
-          reject(err)
+      request({
+        url: url.toString(),
+        headers: {
+          authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
+          'user-agent': 'setup-beam',
+        },
+        fullResponse: false,
+        maxAttempts: 5,
+        retryDelay: 61250,
```
paulo-ferraz-oliveira, line 434: We wait up until 61.25 seconds, on failure. This is because current rate limiting is something like 60 requests per hour. Maybe this is not even important (?)...

src/setup-beam.js
```diff
@@ -413,38 +413,37 @@ function getRunnerOSVersion() {
...
-              }
-            })
-          },
-        )
-        .on('error', (err) => {
-          reject(err)
+      request({
+        url: url.toString(),
+        headers: {
+          authorization: `Bearer ${process.env.GITHUB_TOKEN}`,
```
paulo-ferraz-oliveira, line 429: ⚠️ 👀 ⚠️ 👀 ⚠️ 👀 ⚠️ 👀 

I'm not sure if this (accompanied by the changes in the GitHub workflows) presents any type of security issue, which is why I ask for many eyes on this.

src/setup-beam.js
```diff
@@ -413,38 +413,37 @@ function getRunnerOSVersion() {
...
+          'user-agent': 'setup-beam',
+        },
+        fullResponse: false,
+        maxAttempts: 5,
+        retryDelay: 61250,
+        retryStrategy(err, res) {
+          const retry =
+            !!err || (res.statusCode >= 400 && res.statusCode <= 599)
```
paulo-ferraz-oliveira, line 437: Same strategy as before.

src/setup-beam.js
starbelly: In principle it should be fine, but I don't feel great about it either 🤔

src/setup-beam.js
starbelly: @paulo-ferraz-oliveira WE can use the headers returned to figure out out when to retry. We may just want to use octokit rest package if it already does this though. 

I think we have to have caching to go along with it. I don't like any of these things FWIW 😁

paulo-ferraz-oliveira: Never heard/used Octokit before, but can take a look.

starbelly: It might not support this, dunno. Either way, I think we need to get at doing this : 

https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#conditional-requests

TL;DR we can hit a user's endpoint to see if the rate limit has already been hit, this doesn't count against the rate limit. Thus, we can be sure before we even fire off the first http call to the api for resources that we can actually do it. 

This makes all this more complicated, but it seems worth it. Thoughts?

src/setup-beam.js
kivra-fabber: A fresh GITHUB_TOKEN is created on each workflow run, and expired after it is completed: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

The token does not have permission (by default) to modify the repository, but mostly CI things:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Also note from the GitHub docs:

> Important: An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice, you should always make sure that actions only have the minimum access they require by limiting the permissions granted to the GITHUB_TOKEN.

So the action already theoretically has access to the GITHUB_TOKEN even before explicitly reading it from the process environment. So users depending on actions (and writing their own) should be limiting the scope via CI config anyway, which can be done on at least a per job basis.

kivra-fabber:
Side note from one user perspective: with this patch applied, we have not seen any setup-beam 403 issues in our CI piplines, of which we have run many dozen (some using the action in multiple jobs) since applying. I *believe* the GITHUB_TOKEN usage is the thing that did the trick, so the retries are some kind of extra fallback.

If you want to get some testing done on a branch that only makes changes to use the GITHUB_TOKEN I'd be happy to get that tested. Perhaps it can even use the GitHub context to avoid having to thread the secret through the environment an extra time:
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

paulo-ferraz-oliveira:
@kivra-fabber, thanks for your feedback. It wouldn't hurt to merge, I guess, and expect further feedback from more consumers. We can even remove the retry bits.

kivra-fabber:
If not for any other reasons it would be a simpler fix (just using the GITHUB_TOKEN) to discuss in this PR and merge, and then if we see later that it was not enough we can look at a solution for retrying. If you read it from the GitHub context (`${{ github.token }}`), the fix is even simpler I think.

I think that separate discussion in the original issue, about not even using the API if the user says the version is "locked" (I think?) also is a good thing™️. But can also be a separate fix with its own discussion.

I won't interfere with how you would like to approach this, but I will happily help test and review the PR if it just uses the token. Testing the retries is a bit harder, and like starbelly I think any retry if added should use the timing information in the response headers.

starbelly:
> @kivra-fabber, thanks for your feedback. It wouldn't hurt to merge, I guess, and expect further feedback from more consumers. We can even remove the retry bits.

💯 agree that we should just do the github token part, and then work on retry on a separate PR.

paulo-ferraz-oliveira:
@kivra-fabber, I've updated this to get rid of all non-env.-changed -related code. If you want, give it a go.

kivra-fabber:
@paulo-ferraz-oliveira nice! I'll get it bumped soon in the repo we had issues in. Given that the first iteration of your fix remediated our problems I'm not going to be as fast on it this time, but I expect we can be testing it this week.

paulo-ferraz-oliveira:
@lpil, also tagging you here to see if you wanna test this on top of Gleam. 😁

.github/workflows/action.yml
starbelly: I experienced the same on my current PR. The git version being used on github side maybe changed.

starbelly (approved):

starbelly (changes requested):
@paulo-ferraz-oliveira  I got a 401 when trying this with my fork of gleam. 

https://github.com/starbelly/gleam/actions/runs/3830636416/jobs/6518761952

src/setup-beam.js
```diff
@@ -423,7 +423,10 @@ async function get(url0, pageIdxs) {
         .get(
           url,
           {
-            headers: { 'user-agent': 'setup-beam' },
+            headers: {
+              'user-agent': 'setup-beam',
```
starbelly, line 427: Right, we need get_json function I suppose without the header (seemingly) for the content type a 401 will occur.

starbelly: Upon some investigation it's not the headers, but it consistently fails for windows, all the other envs are fine.

starbelly: Strange it worked at all. I skipped specifying the GITHUB_TOKEN env var in my gleam fork. So, this will _have_ to be optional. That is, we can only conditionally add the header otherwise we'll definitely break what's out there.

src/setup-beam.js
paulo-ferraz-oliveira: No longer valid.

src/setup-beam.js
paulo-ferraz-oliveira: No longer valid.

src/setup-beam.js
paulo-ferraz-oliveira: This is no longer a valid comment.

src/setup-beam.js
paulo-ferraz-oliveira: > Right, we a need get_json function I suppose without the header (seemingly) for the content type a 401 will occur.

I implemented `jsonParse` which should output the non-JSON, the exception, and error out.

starbelly (approved):
Thank you @paulo-ferraz-oliveira !!!

kivra-fabber:
Oh nice, released already 👍 

I _think_ the optional bit inside the JS can be avoided by updating https://github.com/erlef/setup-beam/blob/main/action.yml to pass `${{ github.token }}` into the env of the `runs` script (but I haven't really authored a published GitHub action like this so not sure). As far as I can tell the token is always available there.

starbelly:
> I _think_ the optional bit inside the JS can be avoided by updating https://github.com/erlef/setup-beam/blob/main/action.yml to pass `${{ github.token }}` into the env of the `runs` script (but I haven't really authored a published GitHub action like this so not sure). As far as I can tell the token is always available there.

I'll double check that, if we can avoid people having to modify their workflow, all the better.

paulo-ferraz-oliveira:
> if we can avoid people having to modify their workflow, all the better

I was unaware the token would sometimes not be present, but maybe I overlooked that since I added it to the tests... 😞

starbelly:
@kivra-fabber https://github.com/erlef/setup-beam/pull/172 Thanks for an awesome suggestion 👍

PR #169, opened 2023-01-02
Don't hit github api when an exact version is specified

starbelly:
Recently we've experienced hitting github rate limits when hitting its api to fetch version information from repositories. As a mitigation measure we should avoid hitting their api when an "exact" (i.e., contains major, minor, and patch level parts) version is provided. We can't do this for OTP, since OTP doesn't use semver, however we can avoid hitting github api when version-type strict is true.


src/setup-beam.js
```diff
@@ -321,6 +346,14 @@ function isStrictVersion() {
   return getInput('version-type', false) === 'strict'
 }
 
+function isExactVersion(spec) {
+  if (semver.parse(spec)) {
```
starbelly, line 350: @paulo-ferraz-oliveira  Is there a way to have the linter ignore this function? It's complaining because of useless else, which is somewhat true, but also I don't trust javascript when it comes to boolean logic and `undefined` 😁

starbelly: Ignore this. I went ahead and made a "this is fine" version 😉

paulo-ferraz-oliveira:
@starbelly, `strict` is strict, independent of the source, so I believe OTP should be taken into account, too... but I see you're not going for the option, but rather inferring from the input (as semver), is that it? I think even in that case we could extend not hitting the API for some things (if that's not already done) like listing to find a specific version - if it's not there we error out.

Finally, many tests are failing. Is it possible you use your input on the branch I had already set up to check if the addition of the GitHub secret token doesn't already take care of the initial problem?

src/setup-beam.js
```diff
@@ -321,6 +346,10 @@ function isStrictVersion() {
   return getInput('version-type', false) === 'strict'
 }
 
+function isExactVersion(spec) {
```
paulo-ferraz-oliveira, line 349: I'd call this `isSemver`.
And below I believe you can remove the extra `&& true`.

paulo-ferraz-oliveira:
I believe this one should get merged only after #168. Is that correct, @starbelly?

src/setup-beam.js
paulo-ferraz-oliveira: Right. I was thinking more like

```
return semver.parse(spec) ? true : false
```

(or an `if`) but yours works too, in that case.

It's just that since JavaScript would consider that semver to be a truthy value, the way we're using it in this pull request would still hold, unless it returned `0`, `""`, or `[]`.

It wouldn't if you were doing `isExactVersion(rebar3Spec) == true`.

The joys of JavaScript.

Your solution, though, is more semantically correct.

paulo-ferraz-oliveira: I'd still rename the function for extra clarity, too.

paulo-ferraz-oliveira:
@starbelly, does this one still make sense?

starbelly:
> @starbelly, does this one still make sense?

I think so, if we can mitigate hitting the api, the better off we'll be. A user's github token applies not only to this action but just in general. Thus if you happen have to several workflows across multiple projects, you're going to hit the limit. 

The other part of that is it can be said that hitting an api when you already have all the other information you need is wasteful.

starbelly:
@paulo-ferraz-oliveira  I ripped out the isStrictVersion around OTP as it seems like there's a healthy amount of edge cases, which could be resolved in another PR. I'd rather just mitigate what can be mitigated simply and not shoot for perfection. 

I also moved the logic into the `get*Versions` functions as it seemed tidier.

dist/index.js
```diff
@@ -7350,63 +7350,93 @@ async function getOTPVersion(otpSpec0, osVersion) {
...
-
-  const elixirVersionComp = elixirVersions.get(elixirVersionFromSpec)
-  if (
-    (elixirVersionComp && elixirVersionComp.includes(otpVersionMajor)) ||
-    !isVersion(otpVersionMajor)
-  ) {
-    core.info(
-      `Using Elixir ${elixirVersionFromSpec} (built for Erlang/OTP ${otpVersionMajor})`,
+  let elixirVersionForDownload
+  if (isSemVer(exSpec0)) {
```
starbelly, line 7354: Hmm, this won't actually work well, that is, if we put this here, then we opt of compatibility checks. Maybe we skip doing this for elixir, not sure.

paulo-ferraz-oliveira: Could we maybe go with isStrictVersion, in this case? Or all cases? I'm afraid we're already "guessing" to much. e.g. 0.23 (Gleam) is not SemVer but it is a valid final version, as is Elixir's 1.14, and OTP's 25.1.

paulo-ferraz-oliveira: (I guess this is what's making tests fail for this pull request - ?)

starbelly: I think you are right, and yes, we would have to go with isStrictVersion, though doesn't help @lpil for his case.

starbelly:
Closing this for now.

PR #172, opened 2023-01-06
Provide an input for github tokens

starbelly:
- By providing an input for github tokens with a default users no longer have to pass the github token via their environment across workflows.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,6 +8,15 @@ branding:
   color: blue
   icon: code
 inputs:
+  github_token:
+    description: >
+      A github token used to access the github REST api for retrieving release and version information.
+      While this action will never perform anything more than read optionations against the github REST api,
+      we do not recommend providing a self generated personal access token which may provide more access than is
+      required of this action. This input instead is merely a conveience so that users of this action do not
+      have to manually pass an automatically generated token via their enviroment settings within a workflow.
+      [Learn more about automatically generated tokens](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)
+    default: ${{ github.token }}
   otp-version:
     description: Version range or exact version of Erlang/OTP to use,
       or false when installing only Gleam without OTP
--- a/action.yml
+++ b/action.yml
@@ -14,7 +14,7 @@ inputs:
       While this action will never perform anything more than read optionations against the GitHub REST API,
       we do not recommend providing a self-generated personal access token which may provide more access than is
       required of this action. This input instead is merely a convenience so that users of this action do not
-      have to manually pass an automatically generated token via their enviroment settings within a workflow.
+      have to manually pass an automatically generated token via their environment settings within a workflow.
       [Learn more about automatically generated tokens](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)
     default: ${{ github.token }}
   otp-version:
--- a/action.yml
+++ b/action.yml
@@ -11,7 +11,7 @@ inputs:
   github-token:
     description: >
       A GitHub token used to access the GitHub REST API for retrieving release and version information.
-      While this action will never perform anything more than read optionations against the GitHub REST API,
+      While this action will never perform anything more than read operations against the GitHub REST API,
       we do not recommend providing a self-generated personal access token which may provide more access than is
       required of this action. This input instead is merely a convenience so that users of this action do not
       have to manually pass an automatically generated token via their environment settings within a workflow.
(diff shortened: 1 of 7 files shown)


starbelly:
@paulo-ferraz-oliveira  I also changed the github workflows, for some reason we had the branch main specified, but it's not clear why, and was resulting in failures for pull requests (also not clear why). Was there any particular reason we had it set up like this? Seems odd.

dist/index.js
```diff
@@ -7612,9 +7612,13 @@ async function get(url0, pageIdxs) {
       const headers = {
         'user-agent': 'setup-beam',
       }
-      if (process.env.GITHUB_TOKEN) {
-        headers.authorization = `Bearer ${process.env.GITHUB_TOKEN}`
+      const GithubToken =
+        getInput('github_token', false) || process.env.GITHUB_TOKEN
+
+      if (GithubToken) {
```
starbelly, line 7618: We should probably do a check here to make sure Githubtoken isn't `""`, thoughts?

paulo-ferraz-oliveira:
> but it's not clear why

I usually do it like that, so not odd to me. Whatever pull requests are created they'll only run CI if they target `main`. I guess it doesn't hurt much to not have it, though I don't see what issues it would have caused. On the other hand, if it causes less issues the way you're proposing it now, we can also keep it.

action.yml
```diff
@@ -8,6 +8,15 @@ branding:
   color: blue
   icon: code
 inputs:
+  github-token:
+    description: >
+      A github token used to access the github REST api for retrieving release and version information.
```
paulo-ferraz-oliveira, line 13:
```suggestion
      A GitHub token used to access the GitHub REST API for retrieving release and version information.
```

action.yml
```diff
@@ -8,6 +8,15 @@ branding:
...
+      While this action will never perform anything more than read optionations against the github REST api,
```
paulo-ferraz-oliveira, line 14:
```suggestion
      While this action will never perform anything more than read optionations against the GitHub REST API,
```

paulo-ferraz-oliveira: Also, what's "optionations"?

action.yml
```diff
@@ -8,6 +8,15 @@ branding:
...
+      we do not recommend providing a self generated personal access token which may provide more access than is
```
paulo-ferraz-oliveira, line 15:
```suggestion
      we do not recommend providing a self-generated personal access token which may provide more access than is
```

action.yml
```diff
@@ -8,6 +8,15 @@ branding:
...
+      required of this action. This input instead is merely a conveience so that users of this action do not
```
paulo-ferraz-oliveira, line 16:
```suggestion
      required of this action. This input instead is merely a convenience so that users of this action do not
```

action.yml
```diff
@@ -8,6 +8,15 @@ branding:
...
+      have to manually pass an automatically generated token via their enviroment settings within a workflow.
```
paulo-ferraz-oliveira, line 17:
```suggestion
      have to manually pass an automatically generated token via their environment settings within a workflow.
```

dist/index.js
```diff
@@ -7612,9 +7612,13 @@ async function get(url0, pageIdxs) {
       const headers = {
         'user-agent': 'setup-beam',
       }
-      if (process.env.GITHUB_TOKEN) {
-        headers.authorization = `Bearer ${process.env.GITHUB_TOKEN}`
+      const GithubToken =
```
paulo-ferraz-oliveira, line 7615: I was sure that the linter would complain about non-camelcase, but if it doesn't this is also Ok.

dist/index.js
```diff
@@ -7612,9 +7612,13 @@ async function get(url0, pageIdxs) {
...
+        getInput('github-token', false) || process.env.GITHUB_TOKEN
```
paulo-ferraz-oliveira, line 7616: Why do we need the `||`. It's either defined by the consumer or default, right? Unless you+re thinking the consumer might put a falsy value there.

starbelly: Honestly I'm not sure we need the `or` at all. It depends on whether we want to support providing a token that way. For a second I thought we needed it for backwards compat with the previous release, but since `github.token` is functionally the same, it shouldn't matter (i.e., there will be no breaking change).  Thoughts?

paulo-ferraz-oliveira: I'm Ok with either removing it (not sure it's required) or keeping it (doesn't hurt, but it may be confusing). Your choice.

PR #174, opened 2023-01-07
Simulate input `github-token` to fix unit tests

### LINKED ISSUE #173: Unit tests can fail with 403 (rate limit exceeded)
This problem doesn't effect users of this action, rather they are workflows or part of a workflow that doesn't use our action, and thus do not authenticate with the github REST api. 

Resolution : Find simple and safe way to authenticate with github rest api for npm unit tests.

paulo-ferraz-oliveira:
Fix #173.

Let's hope it's as simple as the first commit I'm pushing.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -8,6 +8,9 @@ on:
   pull_request:
     types: [opened, synchronize]
 
+env:
+  GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
 jobs:
   check_integrity:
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
There might still be some issues with the fact that some tests appear to fail (though some are actually expected to fail, so that's maybe what's confusing me) but I didn't look deep into that.

starbelly (approved):
I love a simple fix!!!

paulo-ferraz-oliveira:
Perfect. Thanks.

PR #183, opened 2023-03-29
Update 3rd party licenses

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
Will do this in a separate pull request to require only single pull request approval.

PR #205, opened 2023-05-09
Eliminate ret in sortVersions

wingyplus:
Use if with return instead of set to `ret`.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10263,7 +10263,6 @@ function maybeCoerced(v) {
 }
 
 function sortVersions(left, right) {
-  let ret = 0
   const newL = verAsComparableStr(left)
   const newR = verAsComparableStr(right)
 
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
Thanks for the pull request. 

When I coded that I wanted to avoid early returns, so the function was easier to reason on (defect of using FP too much? 😆).

Sorry if the code isn't explicit in that sense.

Closing...

wingyplus:
I think declaring `ret` at the beginning of the function is also hard to read. But when I use `if/else if/else`, it got reported by eslint. T_T So I ended up like this.

paulo-ferraz-oliveira:
Not sure how you're running `eslint` but we usually do as:

```
npm install
npm run jslint
```

or

```
npm run build-dist
```

if you wanna run everything required for contributing.

PR #206, opened 2023-05-09
Replace console.log with core.info

wingyplus:


paulo-ferraz-oliveira:
Thanks for this.

This is done in an existing pull request already, though, so I'll close it to avoid merge conflicts.

PR #213, opened 2023-06-15
Fix post -main-merge issue

paulo-ferraz-oliveira:
I committed a last minute change to `setup-beam`, but this caused a CI issue since the same change doesn't show up in `dist/index.js` :(

There's also a minor change to the code where it was identified before in a different pull request (but since that one's not moving much, I decided to do it here).

**Edit**: I've changed the settings so that at least the static analysis elements are compulsory.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -9750,7 +9750,7 @@ async function installOTP(osVersion, otpVersion, hexMirrors) {
       core.addPath(otpPath)
       core.exportVariable('INSTALL_DIR_FOR_OTP', cachedPath)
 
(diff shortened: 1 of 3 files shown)


starbelly (approved):
LGTM!

paulo-ferraz-oliveira:
Many thanks, kind sir.

paulo-ferraz-oliveira:
Release in v1.16.0 / v1.16. v1 tag moved up...

paulo-ferraz-oliveira:

PR #216, opened 2023-06-22
Handle Erlang/OTP versions of type x.y.z.a (non-SemVer)

### LINKED ISSUE #215: Wrong OTP version installed
When using this:

```yaml
env:
  ImageOS: "ubuntu22"
  # Elixir and Erlang versions
  OTP_VERSION: "25.3.2.1"
  ELIXIR_VERSION: "1.14.5"
...
      - name: Setup Erlang, Elixir, etc
        uses: erlef/setup-beam@v1
        with:
          otp-version: ${{ env.OTP_VERSION }}
          elixir-version: ${{ env.ELIXIR_VERSION }}
```

The action outputs:
image

image

Not sure why the OTP version gets changed to `25.3.2.2` when it's `25.3.2.1`.

paulo-ferraz-oliveira:
Fixes #215.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10035,10 +10035,11 @@ async function maybeInstallRebar3(rebar3Spec) {
 async function getOTPVersion(otpSpec0, osVersion, hexMirrors) {
   const otpVersions = await getOTPVersions(osVersion, hexMirrors)
   let otpSpec = otpSpec0 // might be a branch (?)
-  const otpVersion = getVersionFromSpec(
-    otpSpec,
-    Array.from(otpVersions.keys()).sort(),
-  )
+  const otpVersion = getVersionFromSpec({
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira:
We can see it fail (pre-fix) [here](https://github.com/erlef/setup-beam/actions/runs/5345107377/jobs/9690338202?pr=216#step:5:13).

paulo-ferraz-oliveira:
This surfaced another issue in our test assumptions, which is good, and I'll fix :)

paulo-ferraz-oliveira:
Yeah, I experimented a bit with this and don't like the solution I found (and couldn't find a better alternative) or how deep the changes would be to accommodate a very specific version type. Will close this and open a new one (different branch) for some learnings, though.

PR #217, opened 2023-06-23
Rework our version matching algorithm

paulo-ferraz-oliveira:
This:

- bumps dependencies' versions to their latest, where appropriate
- trims GitHub Actions' workflows, where appropriate
- trims `package.json`, where appropriate
- tweaks `getVersionFromSpec/2` to improve on version matching
- adapts test and CI


package.json
```diff
@@ -13,29 +13,20 @@
...
   "devDependencies": {
-    "@vercel/ncc": "0.34.0",
-    "eslint": "8.25.0",
+    "@vercel/ncc": "0.36.1",
+    "eslint": "8.43.0",
     "eslint-config-airbnb": "19.0.4",
-    "eslint-plugin-import": "2.26.0",
-    "eslint-plugin-jsx-a11y": "6.6.1",
-    "eslint-plugin-react": "7.31.10",
-    "husky": "8.0.1",
```
paulo-ferraz-oliveira, line 35: Not really used, so we can trim down a few deps.

src/setup-beam.js
paulo-ferraz-oliveira, line 1: Probably best viewed with "Hide whitespace" enabled.

src/setup-beam.js
```diff
@@ -184,13 +185,16 @@ async function getElixirVersion(exSpec0, otpVersion0, hexMirrors) {
   const otpVersion = otpVersion0.match(/^([^-]+-)?(.+)$/)[2]
   const otpVersionMajor = otpVersion.match(/^([^.]+).*$/)[1]
   const elixirVersions = await getElixirVersions(hexMirrors)
-  const semverVersions = Array.from(elixirVersions.keys()).sort()
```
paulo-ferraz-oliveira, line 187: This is now done inside `getVersionFromSpec/2`, for all consumer functions, as to centralised behaviour and make it more consistent.

src/setup-beam.js
```diff
@@ -221,20 +225,24 @@ async function getElixirVersion(exSpec0, otpVersion0, hexMirrors) {
...
+  const gleamVersion = getVersionFromSpec(spec, versions)
   if (gleamVersion === null) {
     throw new Error(
       `Requested Gleam version (${gleamSpec0}) not found in version list ` +
         "(should you be using option 'version-type': 'strict'?)",
     )
   }
 
-  return maybePrependWithV(gleamVersion, gleamVersion)
+  return maybePrependWithV(gleamVersion)
```
paulo-ferraz-oliveira, line 238: The 2nd argument had no value, as it wasn't accepted, in any case.

src/setup-beam.js
```diff
@@ -300,7 +309,7 @@ async function getElixirVersions(hexMirrors) {
     .forEach((line) => {
       const elixirMatch =
         line.match(/^v?(.+)-otp-([^ ]+)/) || line.match(/^v?([^ ]+)/)
-      const elixirVersion = maybePrependWithV(elixirMatch[1])
```
paulo-ferraz-oliveira, line 303: This is no longer required, since internally all versions will be stripped of `v` and it'll only be added when required for external requests.

src/setup-beam.js
```diff
@@ -333,62 +342,63 @@ async function getRebar3Versions() {
...
   })
+
   return rebar3VersionsListing
 }
 
 function isStrictVersion() {
   return getInput('version-type', false) === 'strict'
 }
 
-function getVersionFromSpec(spec, versions, maybePrependWithV0) {
```
paulo-ferraz-oliveira, line 350: Most of the big changes are in this function... and `maybeCoerced/1` below.

src/setup-beam.js
```diff
@@ -333,62 +342,63 @@ async function getRebar3Versions() {
...
+        `you have to set version-type=strict when using a semver-invalid version (got ${spec0})`,
+      )
+    }
   }
 
-  return v
+  return version || null
 }
 
 function maybeCoerced(v) {
```
paulo-ferraz-oliveira, line 394: Most of the big changes are in this function... and `getVersionFromSpec/2` above.

test/setup-beam.test.js
paulo-ferraz-oliveira, line 1: In this, we basically show that if it's not `strict` on input, we expected a SemVer output, to allow the input coercion to tie properly with the internal versions' map.

starbelly:
> It is my experience that it's rare the development community at large targets versions like 25.3.2.1, preferring to stay at the 25.3.x level.

Yeah, I agree with that, but the fact remains still that even 25.3.1 is not semver in the case of OTP. However, I think it's fair to say it can be treated like that. I would agree this is a breaking change. So, we're thinking this would result in a new major version then, is that correct?

paulo-ferraz-oliveira:
It'd probably result in a breaking change, yes, depending on:

* whether we consider it to be a bug (which it is since we've not been handling it properly), or
* we consider it to be a new _feature_ we want the action to present.

I "think" I can try to fit the `x.y.z.a` versions in the mix (maybe?), but I tried the other day and the solution got so convoluted I rolled back some decisions: I also don't like that I'm forced to write `25.3.2.2, strict`, because my consumer's expectation is 25.3.2 is whatever is before `25.4.0`, but... Now I got (merge) conflicts, I got solve those and we can continue discussing this to see what our common expectation is.

paulo-ferraz-oliveira:
I rebased to resolve merge conflict and increased complexity for `getVersionFromSpec/2` (hopefully not by much) so we can keep previous behaviour but make the code easier to maintain, also...

src/setup-beam.js
```diff
@@ -162,43 +163,41 @@ async function maybeInstallRebar3(rebar3Spec) {
...
 async function getOTPVersion(otpSpec0, osVersion, hexMirrors) {
   const otpVersions = await getOTPVersions(osVersion, hexMirrors)
-  let otpSpec = otpSpec0 // might be a branch (?)
-  const otpVersion = getVersionFromSpec(
-    otpSpec,
-    Array.from(otpVersions.keys()).sort(),
-  )
-  if (isVersion(otpSpec0)) {
-    otpSpec = `OTP-${otpSpec0}` // ... it's a version!
-  }
```
paulo-ferraz-oliveira, line 172: The result of this block was never used. I don't know why static analysis wouldn't find this...

src/setup-beam.js
```diff
@@ -483,31 +491,53 @@ async function get(url0, pageIdxs) {
...
+
+function xyzabVer(pref, suf) {
+  // This accounts for stuff like 6.0.2.0.1.0, as proposed by Erlang's
+  // https://www.erlang.org/doc/system_principles/versions.html
+  const dd = '\\.?(\\d+)?'
+  return new RegExp(
+    `${pref}v?(\\d+)${dd}${dd}${dd}${dd}${dd}(?:-rc\\.?\\d+)?(?:-otp-\\d+)?${suf}`,
+  )
+}
+
```
paulo-ferraz-oliveira, line 538: This function, and it's callers, will probably be the "most difficult bits" to maintain, but the regexps here are pretty relaxed, too...

paulo-ferraz-oliveira:
@starbelly?

src/setup-beam.js
starbelly: Needs hank.

starbelly (approved):
@paulo-ferraz-oliveira This LGTM. I'd be in favor of merging and a delay on the release, sound good?

paulo-ferraz-oliveira:
Sure. We can merge and delay release. That'll give me a chance to update the other pull request I have opened.

paulo-ferraz-oliveira:
Force-pushing to resolve merge conflicts. If no actions show issues I'm squash-merging.

josevalim:
Thank you @paulo-ferraz-oliveira! It seems this PR removed support for maint/master from OTP git, which Elixir relies on. Was it intentional?

paulo-ferraz-oliveira:
> It seems this PR removed support for maint/master from OTP git, which Elixir relies on. Was it intentional?

👋 It wasn't. I'll look at the pull request.

PR #228, opened 2023-08-15
Add ref: (version) support to .tool-versions parsing

dvic:
# Description

This adds support for using source code elixir references in `.tool-versions`, e.g., `elixir ref:v1.15.4`.

- [x] I have performed a self-review of my changes (*I was not able to run the tests locally on my mac though)
- [x] I have read and understood the [contributing guidelines](https://github.com/erlef/setup-beam/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -602,11 +602,11 @@ function parseVersionFile(versionFilePath0) {
   // If we ever start parsing something else, this should
   // become default in a new option named e.g. version-file-type
   versions.split('\n').forEach((line) => {
-    const appVersion = line.match(/^([^ ]+)[ ]+([^ #]+)/)
+    const appVersion = line.match(/^([^ ]+)[ ]+(ref:v)?([^ #]+)/)
     if (appVersion) {
       const app = appVersion[1]
       if (['erlang', 'elixir', 'gleam', 'rebar'].includes(app)) {
-        const [, , version] = appVersion
+        const [, , , version] = appVersion
         core.info(`Consuming ${app} at version ${version}`)
         appVersions.set(app, version)
       }
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -602,7 +602,7 @@ function parseVersionFile(versionFilePath0) {
   // If we ever start parsing something else, this should
   // become default in a new option named e.g. version-file-type
   versions.split('\n').forEach((line) => {
-    const appVersion = line.match(/^([^ ]+)[ ]+(ref:v)?([^ #]+)/)
+    const appVersion = line.match(/^([^ ]+)[ ]+(ref:v?)?([^ #]+)/)
     if (appVersion) {
       const app = appVersion[1]
       if (['erlang', 'elixir', 'gleam', 'rebar'].includes(app)) {
--- a/test/setup-beam.test.js
+++ b/test/setup-beam.test.js
@@ -510,16 +510,16 @@ async function testParseVersionFile() {
   const elixir = '1.14.1'
   const gleam = '0.23.0'
   const rebar3 = '3.16.0'
-  let toolVersions = `# a comment
-erlang   ${erlang}# comment, no space
-elixir ${elixir}  # comment, with space
+  const toolVersions = `# a comment
+erlang   ref:v${erlang}# comment, no space, and ref:v
+elixir ref:${elixir}  # comment, with space and ref:
  not-gleam 0.23 # not picked up
 gleam ${gleam} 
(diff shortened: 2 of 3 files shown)


dvic:
Using this syntax in your `.tool-versions` will compile elixir from source, the benefit for me is that I have goto definition for all Elixir source code in my editor.

paulo-ferraz-oliveira:
Thanks for this. I'll comment on the changes, where appropriate.

src/setup-beam.js
```diff
@@ -602,11 +602,11 @@ function parseVersionFile(versionFilePath0) {
   // If we ever start parsing something else, this should
   // become default in a new option named e.g. version-file-type
   versions.split('\n').forEach((line) => {
-    const appVersion = line.match(/^([^ ]+)[ ]+([^ #]+)/)
+    const appVersion = line.match(/^([^ ]+)[ ]+(ref:v)?([^ #]+)/)
```
paulo-ferraz-oliveira, line 605: The `v` should probably be optional, right?

Other `ref:` might be `ref:39cb398vb39` as explained in [Configuration | asdf](https://asdf-vm.com/manage/configuration.html#tool-versions).

test/setup-beam.test.js
```diff
@@ -540,6 +540,13 @@ rebar ${rebar3}`
   simulateInput('elixir-version', elixirVersion)
   simulateInput('gleam-version', gleamVersion)
   simulateInput('rebar3-version', rebar3Version)
+
+  // test that elixir ref:v syntax works
+  toolVersions = `erlang ${erlang}\nelixr ref:v${elixir}`
+  fs.writeFileSync(filename, toolVersions)
+  appVersions = setupBeam.parseVersionFile(filename)
+  assert.strictEqual(appVersions.get('erlang'), erlang)
+  assert.strictEqual(appVersions.get('elixir'), elixir)
```
paulo-ferraz-oliveira, line 549: It would be preferable that you changed the `toolVersions` constant, above, instead of adding to it and rewriting the file, unless there's a good reason for this I'm not seeing. Also, it seems there's a typo: `elixr` instead of `elixir`.

paulo-ferraz-oliveira: You can replace `elixir ${elixir}` with `elixir ref:${elixir}` (no `v`), no problem. This way we test the optional `v` and your changes in a single go, what do you say?

dvic: but then we don't test that it also works without the `ref:` right?

paulo-ferraz-oliveira:
Tests are failing, most likely for:

* `using elixr` instead of `elixir`, which will return `undefined` when trying to get a version
* not having run `npm run build-dist` (as per the [contributing guidelines](https://github.com/erlef/setup-beam/blob/main/CONTRIBUTING.md)).

test/setup-beam.test.js
paulo-ferraz-oliveira: That constant is

```
erlang   ${erlang}# comment, no space
elixir ${elixir}  # comment, with space
 not-gleam 0.23 # not picked up
gleam ${gleam} 
rebar ${rebar3}`
```

so there's tests for "with `ref:`" and "without `ref:`".

dvic:
Should I add an issue for documentation? By the way the contributing link is broken in the template: 

`https://github.com/erlef/setup-beam/blob/main/erlef/setup-beam/blob/main/CONTRIBUTING.md`, it should be
`https://github.com/erlef/setup-beam/blob/main/CONTRIBUTING.md`

paulo-ferraz-oliveira:
There's no need to document :) you're just implementing the standard, so whoever's using `asdf` should know (if they need it).

As for the "contributing" link, thanks for noticing. I'll fix it in a subsequent pull request.

paulo-ferraz-oliveira:
All checks successful, will merge. Thanks for this. It shall be present on the next release, hopefully soonish.

paulo-ferraz-oliveira (approved):

dvic:
> **Edit**: the "contributing" link is set as `/erlef/setup-beam/blob/main/CONTRIBUTING.md`, so I don't understand why it'd be broken for you.

Now it works, I guess a github bug? 🤷🏻‍♂️

dvic:
> All checks successful, will merge. Thanks for this. It shall be present on the next release, hopefully soonish.

Any chance to cut a release ^^?

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/releases/tag/v1.17.0
v1.17 is the same commit
v1 was moved to the latest commit

PR #229, opened 2023-08-24
Ease pre-pull request fork CI as well as github-actions based dispatch

paulo-ferraz-oliveira:
# Description

We allow GitHub actions pull requests to dispatch the other workflows when version updates are in order.

We also simplify the other conditions to be simply "on push".

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,13 +1,7 @@
 ---
 name: action
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+on: [push, workflow_dispatch]
 
 env:
   GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
@starbelly, don't really understand why CI isn't running. It's either basing itself on the new conditions (which are "all push", so should work) or the previous conditions (which included "pull request to main", so should work). I believe you can "trust it", and approve+merge. We can later see if it runs in `master` or not, and try to bugfix it, if not.

starbelly:
@paulo-ferraz-oliveira I've experienced this before :( Nothing to do with this action AFAIK. What you can do is push an empty commit to give it a jiggle as FYI.

starbelly (approved):
I'm happy to trust approve this one 😄 There must be _some_ trust in this world ❤️

paulo-ferraz-oliveira:
There it goes, queue in `master`

image

❤️

PR #236, opened 2023-10-23
update checkout to v4

lordB8r:
# Description

Updated checkout to most current version (v4)

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -11,7 +11,7 @@ jobs:
     name: Expected local npm actions
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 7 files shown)


starbelly:
Not sure why this isn't running, also need to understand what v4 gives us and issues with v3.

paulo-ferraz-oliveira:
I'm also curious about the need for this change... I don't see any reports on security-related issues, or deprecations.

paulo-ferraz-oliveira (changes requested):
Check the comments above.

paulo-ferraz-oliveira:
I'm closing without merging for lack of response and because we don't see the need to do this, at this moment.

PR #239, opened 2024-01-11
Have the action output a versioning element so it could be used as part of the cache restore key

### LINKED ISSUE #222: Have the action output a versioning element so it could be used as part of the cache restore key
When we release a new version, like 1.16 (that "broke" cached PLTs), this could help avoid some issues on consumers.

paulo-ferraz-oliveira:
# Description

We implement auto-versioning (from the main branch) as a way to have an action output variable which is the commit unique id of the action version being executed, and also document this.

Since this implies a change to the main branch, directly, and there's no easier way to do it, we'd have to change

image

(which I already did, so needs to be reverted if we don't wanna move forward 😄)

These (changes to the settings) do not represent **explicit** security issues, only potential maintenance issues, in which we'd merge something that didn't pass CI. In any case, I'm "pretty sure" that before releasing, and especially at pr time, we validate the workflow results and ask pr'ers to act on those, so it shouldn't be a big deal. Lemme know...

Otherwise, if you have other ideas (I'd hate to have to approve a pull request for this everytime we merge to the main branch or, even worse, have to maintain the version by hand), I'm all ears...

Closes #222.

## Other considerations

There's a bit of linting and clean up involved, also.

And finally, I tested the changes and the flow in my own action branch, so it should be working as-is.


starbelly (approved):
LGTM!

PR #240, opened 2024-01-13
 Bump semver to 7.5.4

paulo-ferraz-oliveira:
# Description

... as per Dependabot ([CVE-2022-25883](https://nvd.nist.gov/vuln/detail/CVE-2022-25883)).

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -19,15 +19,18 @@ jobs:
           SHA=$(git rev-parse --short HEAD)
           sed -i'.bak' \
               -e "s/const setupBeamVersion = '.*'/const setupBeamVersion = '${SHA}'/g" \
-              dist/index.js
+              src/setup-beam.js
+          npm run build-dist
           git config user.name "GitHub Actions"
           git config user.email "actions@user.noreply.github.com"
+          git add src/setup-beam.js
           git add dist/index.js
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
We should probably create a minor release, after this one, for the security update, and for the new feature.

.github/workflows/action.yml
```diff
@@ -20,13 +20,6 @@ jobs:
           sed -i'.bak' \
               -e "s/const setupBeamVersion = '.*'/const setupBeamVersion = '${SHA}'/g" \
               dist/index.js
-          # it always starts by being 'undefined'
```
starbelly, line 23: Is this intentional?

paulo-ferraz-oliveira: Yes. I mean it's not directly related to my initial intention, but it doesn't always start by being `undefined`, since once we change it, it goes to the `dist` and then it's not `undefined` any more 👍

paulo-ferraz-oliveira: I ended up tweaking it more after I tried to do another pull request and run into "issues". (basically we need to update the `.js` that we usually do and then run `npm run build-dist` instead of only doing the change in the `dist` files)

starbelly (approved):

PR #242, opened 2024-01-17
Add known branch exceptions to version matching

paulo-ferraz-oliveira:
# Description

Maybe doing it this way is better, since we don't "break" existing consumption elements, while just adding to already existing exception `-rc-`.

Forcing `version-type: strict`, while explicit, would probably force some installations to use a too-specific Elixir version, so we "fix" what we had to prevent that.

Closes #241.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -10224,7 +10224,7 @@ function getVersionFromSpec(spec0, versions0) {
   const rangeMax = semver.maxSatisfying(versions, rangeForMax)
   let version = null
 
-  if (isStrictVersion() || isRC(spec0)) {
+  if (isStrictVersion() || isRC(spec0) || isKnownBranch(spec0)) {
     if (versions0[spec]) {
       // If `version-type: strict` or version is RC, we obtain it directly
       version = versions0[spec]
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira:
@josevalim, @grzuy, this adds an exception (to loose version matching) for "known" branches `main`, `maint`, and `master`. Would you be so kind as to check the tests and maybe run this branch where it's failing now, so we can move to merge it? Thanks.

dist/index.js
```diff
@@ -10292,6 +10292,10 @@ function isRC(ver) {
   return ver.match(xyzAbcVersion('^', '(?:-rc\\.?\\d+)'))
 }
 
+function isKnownBranch(ver) {
+  return ['main', 'master', 'maint'].includes(ver)
```
josevalim, line 10296: If you want to be safer, you could match "maint" anywhere in the version, as OTP has maint-26, maint-25, and so on. :) Just saying, your call, we don't use it right now.

josevalim:
Looks great, thank you!

starbelly (approved):
LGTM!

paulo-ferraz-oliveira:
Perfect stuff. Thanks @starbelly. I'll release a patch now.

paulo-ferraz-oliveira:
v1, v1.17 and v1.17.1 released, pointing to the latest commit.

PR #244, opened 2024-01-25
Update outdated/deprecated actions

paulo-ferraz-oliveira:
# Description

- moves all actions/checkout to @v4
- moves all erlef/setup-beam to @v1
- moves all actions/setup-node to @v4

(this also includes what we suggest in README)

Closes #243.

- [x] I have performed a self-review of my changes
- [x] I have read and understood the [contributing guidelines](/erlef/setup-beam/blob/main/CONTRIBUTING.md)

PR #246, opened 2024-01-28
Fix "Windows workflow broken since recent commit"

paulo-ferraz-oliveira:
# Description

Closes #244. Check that for more information.

CI now proves the implementation of the environment variables is Ok, as was expected.


.github/workflows/ubuntu.yml
paulo-ferraz-oliveira, line 1: Instead of "not failing it they're not there" (this was a best effort, after all), we now fail if the variables are incorrectly populated. This is Ubuntu!

.github/workflows/windows.yml
paulo-ferraz-oliveira, line 1: Instead of "not failing it they're not there" (this was a best effort, after all), we now fail if the variables are incorrectly populated. This is Windows.

dist/.github/elixir-matchers.json
paulo-ferraz-oliveira, line 1: All the removed `dist` files are no longer required. They're remains from when the action did stuff using shell scripts, which it no longer does. The `npm` stuff's now been fixed so `dist` is always recreated.

dist/index.js
paulo-ferraz-oliveira, line 1: Changes here:
1. fix broken variable path to include `bin`
2. fix Windows/rebar3 location to have code similar to Ubuntu, on Windows (also inside `bin`

dist/install-gleam.ps1
paulo-ferraz-oliveira, line 1: All the removed `dist` files are no longer required. They're remains from when the action did stuff using shell scripts, which it no longer does. The `npm` stuff's now been fixed so `dist` is always recreated.

dist/install-gleam.sh
paulo-ferraz-oliveira, line 1: All the removed `dist` files are no longer required. They're remains from when the action did stuff using shell scripts, which it no longer does. The `npm` stuff's now been fixed so `dist` is always recreated.

dist/install-rebar3.ps1
paulo-ferraz-oliveira, line 1: All the removed `dist` files are no longer required. They're remains from when the action did stuff using shell scripts, which it no longer does. The `npm` stuff's now been fixed so `dist` is always recreated.

dist/install-rebar3.sh
paulo-ferraz-oliveira, line 1: All the removed `dist` files are no longer required. They're remains from when the action did stuff using shell scripts, which it no longer does. The `npm` stuff's now been fixed so `dist` is always recreated.

src/setup-beam.js
paulo-ferraz-oliveira, line 1: Check comments in the `dist/.js` file.

starbelly (approved):

PR #247, opened 2024-01-28
Have `npm run build-dist` _always_ recreate `dist`

paulo-ferraz-oliveira:
# Description

So we don't get keep files we no longer need. I wanted to have pushed this with #246 but I didn't push the local commit 😄 


paulo-ferraz-oliveira:
No hurry on this one. 😄

starbelly (approved):

PR #252, opened 2024-02-03
Support for mise / .mise.toml version files

### LINKED ISSUE #250: Support for mise (previously known as rtx)
Currently, setup-beam supports asdf's `.tool-versions` file.

A similar tool that is increasing in popularity is [mise](https://mise.jdx.dev/), previously known as rtx but recently renamed because of the naming conflict with NVIDIA's RTX graphics cards.

mise uses a `.mise.toml` file which fulfils a similar role to `.tool-versions`.  Here's an example of what one looks like:

```toml
[env]
_.file = '.env'
KERL_BUILD_DOCS="yes"

[tools]
elixir = "1.16.0-otp-26"
erlang = "26.2.1"
```

I'm happy to write a PR to support mise, but wanted to check in with the maintainers to see that it would be a valuable and accepted contribution first.

mogest:
# Description

setup-beam currently supports the asdf version file format.  This PR extends it to also support the [mise](https://mise.jdx.dev/) (previously known as rtx) format as well.

A new option, `version-file-type` has been added which takes the values `asdf` or `mise`.  If it's not specified, it defaults to `asdf` (for backwards compatibility) and warns the user that it is defaulting.

A TOML parsing library has been included with the build to read the .mise.toml file.  @iarna/toml was selected due to its relatively small unpacked file size and modern support.

For more details, the implementation strategy is laid out in issue #250.

Resolves #250

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -55,6 +55,14 @@ jobs:
         uses: ./
         with:
           version-file: test/.tool-versions
+          version-file-type: asdf
+          version-type: strict
+      - run: rm -rf /opt/hostedtoolcache
+      - name: .mise.toml test
+        uses: ./
+        with:
+          version-file: test/.mise.toml
+          version-file-type: mise
           version-type: strict
 
   unit_tests_windows:
--- a/README.md
+++ b/README.md
@@ -105,8 +105,10 @@ e.g. `${{steps.setup-beam.outputs.erlang-version}}`
 
 ### Version file
 
-A version file is specified via input `version-file` (e.g.`.tool-versions`). This
-allows not having to use YML input for versions, though the action does check (and
+A version file can be used via inputs `version-file` which specifies the file path (e.g.`.tool-versions`)
+and `version-file-type` which can be set to either `asdf` or `mise`.
+
+This allows not having to use YML input for versions, though the action does check (and
 will exit with error) if both inputs are set.
 
 **Note**: if you're using a version file, option `version-type` is checked to be `strict`,
@@ -115,18 +117,19 @@ and will make the action exit with error otherwise.
 The following version file formats are supported:
 
 - `.tool-versions`, as specified by [asdf: Configuration](https://asdf-vm.com/manage/configuration.html)
+- `.mise.toml`, as specified by [mise: Configuration](https://mise.jdx.dev/configuration.html)
 
 Supported version elements are the same as the ones defined for the YML portion of the action,
 with the following correspondence.
 
-#### `.tool-versions` format
+#### Mapping between YML and version file formats
 
-| YML              | `.tool-versions` |
-|-                 |-
-| `otp-version`    | `erlang`
-| `elixir-version` | `elixir`
-| `gleam-version`  | `gleam`
-| `rebar3-version` | `rebar`
+| YML              | `.tool-versions` | `.mise.toml`
+|-                 |-                 |-
+| `otp-version`    | `erlang`         | `erlang`
+| `elixir-version` | `elixir`         | `elixir`
+| `gleam-version`  | `gleam`          | `gleam`
+| `rebar3-version` | `rebar`          | `rebar`
 
 ### Example (Erlang/OTP + Elixir, on Ubuntu)
 
--- a/action.yml
+++ b/action.yml
@@ -43,7 +43,10 @@ inputs:
       (or not)
     default: false
   version-file:
-    description: a versions file (e.g. as used by `asdf`), which defines inputs
+    description: a versions file (e.g. as used by `asdf` or `mise`), which defines inputs
+    default: ''
+  version-file-type:
+    description: which version file type to use, supported values are `asdf` and `mise`
     default: ''
   hexpm-mirrors:
     description: mirror(s) for hex.pm, one per line
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "@actions/exec": "1.1.1",
         "@actions/http-client": "2.1.0",
         "@actions/tool-cache": "2.0.1",
+        "@iarna/toml": "^2.2.5",
         "semver": "7.5.4"
       },
       "devDependencies": {
@@ -174,6 +175,11 @@
       "resolved": "https://registry.npmjs.org/@humanwhocodes/object-schema/-/object-schema-1.2.1.tgz",
       "integrity": "sha512-ZnQMnLV4e7hDlUvw8H+U8ASL02SS2Gn6+9Ac3wGGLIe7+je2AeAOxPY+izIPJDfFDb7eDjev0Us8MO1iFRN8hA==",
       "dev": true
+    },
+    "node_modules/@iarna/toml": {
+      "version": "2.2.5",
+      "resolved": "https://registry.npmjs.org/@iarna/toml/-/toml-2.2.5.tgz",
+      "integrity": "sha512-trnsAYxU3xnS1gPHPyU961coFyLkh4gAD/0zQ5mymY4yOZ+CYvsPqUbOFSw0aDM4y0tV7tiFxL/1XfXPNC6IPg=="
     },
     "node_modules/@isaacs/cliui": {
       "version": "8.0.2",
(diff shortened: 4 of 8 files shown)


.github/workflows/action.yml
```diff
@@ -55,6 +55,14 @@ jobs:
         uses: ./
         with:
           version-file: test/.tool-versions
+          version-file-type: asdf
+          version-type: strict
+      - run: rm -rf /opt/hostedtoolcache
```
paulo-ferraz-oliveira, line 60: Is this `rm -rf` required? Why?

.github/workflows/action.yml
```diff
@@ -55,6 +55,14 @@ jobs:
...
+      - name: .mise.toml test
+        uses: ./
+        with:
+          version-file: test/.mise.toml
```
paulo-ferraz-oliveira, line 64: This file should be added to `.gitignore`.

README.md
```diff
@@ -105,8 +105,10 @@ e.g. `${{steps.setup-beam.outputs.erlang-version}}`
 
 ### Version file
 
-A version file is specified via input `version-file` (e.g.`.tool-versions`). This
-allows not having to use YML input for versions, though the action does check (and
+A version file can be used via inputs `version-file` which specifies the file path (e.g.`.tool-versions`)
+and `version-file-type` which can be set to either `asdf` or `mise`.
```
paulo-ferraz-oliveira, line 109: I'd prefer "... and `version-file-type` which can be set to one of the values in table _Mapping between YML and version file formats_" (where there'd be a link to the table).

README.md
```diff
@@ -105,8 +105,10 @@ e.g. `${{steps.setup-beam.outputs.erlang-version}}`
...
+
+This allows not having to use YML input for versions, though the action does check (and
 will exit with error) if both inputs are set.
```
paulo-ferraz-oliveira, line 112: You probably need to rephrase to accommodate for the newer option.

README.md
```diff
@@ -115,18 +117,19 @@ and will make the action exit with error otherwise.
 The following version file formats are supported:
 
 - `.tool-versions`, as specified by [asdf: Configuration](https://asdf-vm.com/manage/configuration.html)
+- `.mise.toml`, as specified by [mise: Configuration](https://mise.jdx.dev/configuration.html)
 
 Supported version elements are the same as the ones defined for the YML portion of the action,
 with the following correspondence.
 
-#### `.tool-versions` format
+#### Mapping between YML and version file formats
```
paulo-ferraz-oliveira, line 125: This should be "Mapping between YML and `version-file-type` formats", where we list `asdf` and `mise` instead of the file names (which can be chosen by the consumer).

paulo-ferraz-oliveira:
You need to update `action.yml` (at the root folder).

action.yml
```diff
@@ -43,7 +43,10 @@ inputs:
       (or not)
     default: false
   version-file:
-    description: a versions file (e.g. as used by `asdf`), which defines inputs
+    description: a versions file (e.g. as used by `asdf` or `mise`), which defines inputs
```
paulo-ferraz-oliveira, line 46: We don't need the "or `mise`" in this context.

action.yml
```diff
@@ -43,7 +43,10 @@ inputs:
...
+    default: ''
+  version-file-type:
+    description: which version file type to use, supported values are `asdf` and `mise`
```
paulo-ferraz-oliveira, line 49:
```suggestion
    description: version file type to use; supported values are `asdf` and `mise`
```

package-lock.json
```diff
@@ -11,6 +11,7 @@
         "@actions/exec": "1.1.1",
         "@actions/http-client": "2.1.0",
         "@actions/tool-cache": "2.0.1",
+        "@iarna/toml": "^2.2.5",
```
paulo-ferraz-oliveira, line 14: Let's remove the `^`. We want results to be predictable in both consumption and our own CI.

package.json
```diff
@@ -18,6 +18,7 @@
     "@actions/exec": "1.1.1",
     "@actions/http-client": "2.1.0",
     "@actions/tool-cache": "2.0.1",
+    "@iarna/toml": "^2.2.5",
```
paulo-ferraz-oliveira, line 21: Let's remove the `^`. We want results to be predictable in both consumption and our own CI.

src/setup-beam.js
```diff
@@ -49,7 +50,7 @@ async function main() {
   await maybeInstallRebar3(rebar3Spec)
 
   // undefined is replaced by a function, post- main branch merge
-  const setupBeamVersion = '6c8deb6'
+  const setupBeamVersion = 'b80b314'
```
paulo-ferraz-oliveira, line 53: It's strange you'd do this change, so it's probably coming from not using the most recent `main` branch. You should probably rebase.

src/setup-beam.js
```diff
@@ -2,6 +2,7 @@ const core = require('@actions/core')
 const { exec } = require('@actions/exec')
 const tc = require('@actions/tool-cache')
 const http = require('@actions/http-client')
+const TOML = require('@iarna/toml')
```
paulo-ferraz-oliveira, line 5: Let's name this `toml`.

src/setup-beam.js
```diff
@@ -603,13 +604,45 @@ function parseVersionFile(versionFilePath0) {
     )
   }
 
-  core.startGroup(`Parsing version file at ${versionFilePath0}`)
+  let type = getInput('version-file-type', false)
+  if (!type) {
+    type = 'asdf'
+    core.warning("version-file-type not set, defaulting to 'asdf'")
```
paulo-ferraz-oliveira, line 610:
```suggestion
    core.warning("version-file-type not set; defaulting to 'asdf'")
```

src/setup-beam.js
```diff
@@ -603,13 +604,45 @@ function parseVersionFile(versionFilePath0) {
...
+  }
+
+  core.startGroup(`Parsing ${type} version file at ${versionFilePath0}`)
```
paulo-ferraz-oliveira, line 613:
```suggestion
  core.startGroup(`Parsing ${type}-typed version file at ${versionFilePath0}`)
```

src/setup-beam.js
```diff
@@ -620,13 +653,26 @@ function parseVersionFile(versionFilePath0) {
...
-  core.endGroup()
+  return appVersions
+}
 
+function parseMiseVersionFile(fileData) {
+  const appVersions = new Map()
+  const toml = TOML.parse(fileData)
+  for (let app of ['erlang', 'elixir', 'gleam', 'rebar']) {
+    let version = toml.tools && toml.tools[app]
+    if (Array.isArray(version)) {
```
paulo-ferraz-oliveira, line 664: I'd add a comment here: "In case of multiple versions, only the first one is picked up".

src/setup-beam.js
```diff
@@ -620,13 +653,26 @@ function parseVersionFile(versionFilePath0) {
...
+      version = version[0]
+    }
+    if (typeof version === 'object' && version.version) {
```
paulo-ferraz-oliveira, line 667: I'd add a comment here: "In case of a version object, we look for a specific key - version".

paulo-ferraz-oliveira: Also, is this documented in the `mise` format? It'd be nice to add a comment (PR only, not code) here, for future reference.

paulo-ferraz-oliveira (changes requested):
👋 
This looks good. I've requested a few changes, but they should be minor.
I don't understand why the actions aren't running but e.g. if you don't have the TOML file to `.gitignore` it'll most likely fail when I understand how to force them.

paulo-ferraz-oliveira:
@mogest, do you still think you'll tackle this. It's Ok to close temporarily until you get around to it, if you want. In any case, as of this moment, you'd have to rebase since you've diverged from the main branch.

paulo-ferraz-oliveira:
@mogest, this was left unhandled, and got stale due to several releases in the meantime. At this moment, it's got a bunch of merge conflicts that need resolving. I'm Close'ing as stale, with this comment. Feel free to re-open in case you wanna complete the previous changes. 👍

PR #274, opened 2024-05-21
Test over OTP 27 and rebar3 3.23

paulo-ferraz-oliveira:
# Description

I make explicit tests over the newest version of `rebar3` over the newest version of Erlang/OTP. Can't add Elixir to the mix, still, but once we can (1.16, 1.17?) the file is easily updated.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -23,6 +23,7 @@ jobs:
       matrix:
         combo:
           - otp-version: '27'
+            elixir-version: 'v1.17.0-rc1'
             rebar3-version: '3.23'
             os: 'ubuntu-latest'
           - otp-version: '26.0'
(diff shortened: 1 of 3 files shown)


ericmj (approved):
Elixir 1.17 RC will be released soon which you can test with OTP 27.

paulo-ferraz-oliveira:
> Elixir 1.17 RC will be released soon which you can test with OTP 27.

Cool, this can wait then, since I don't think we're in a hurry. I was just trying to assess how ready the action was 😄 I'll prepare the code for 1.17 rc1, in that case. Thanks.

paulo-ferraz-oliveira:
@ericmj, I'm merging with minor changes to the README (if CI passes), which we can easily patch if they're wrong.

I thought it appropriate to review that file in the scope of this PR.

I also did two minor fixes to the tests: one was failing, one would fail in a year or so, when support for latest Elixir would drop support for OTP 25.

PR #281, opened 2024-06-12
Test under non-RC Elixir 1.17

paulo-ferraz-oliveira:
# Description

Not that non-RC Elixir 1.17 is release, we use that in our tests.


paulo-ferraz-oliveira:
If CI's Ok, I'm merging this. Reviewers will have knowledge of the change, but this doesn't affect the solution, only the way it's tested (by updating the tests).

PR #287, opened 2024-06-16
Fixup action name for the GH marketplace

starbelly:
We seemed to have published to the marketplace previously with a description of the action vs the action name. This commit changes the name to simply `setup-beam`.

The link that is out there : `https://github.com/marketplace/actions/setup-erlang-otp-with-optional-elixir-and-mix-and-or-rebar3` 

Additionally, if you search for `setup-beam` or `beam` this will action not come up. We may want to revise the description while we are here. 

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,5 +1,5 @@
 ---
-name: Setup Erlang/OTP with optional Elixir (and mix) and/or rebar3
+name: setup-beam
 description: >
   Set up a specific version of Erlang/OTP, Elixir, and/or rebar3
   and add the command-line tools to the PATH


paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
I did the first change to the name, and at that time I didn't look at the `action.yml` reference. The description was previously "Setup Elixir" 😄

paulo-ferraz-oliveira:
Something like

```
  This action's goal is to set up a BEAM-based CI environment, with
  Erlang/OTP as a base, then optionally, Elixir, Gleam, `rebar3` or `hex`
  on top of that
```

?

I'm also Ok merging this now and changing the description in a subsequent pull request.

paulo-ferraz-oliveira:
Bryan, I'm merging this to check if the verification process moves forward, and then I'll pr to fix the description 😄

PR #294, opened 2024-06-22
Reduce action description in action.yml (GHA Marketplace constraint)

starbelly:
# Description

A brief description of your changes.

Github marketplace requires a description be no longer than 125 chars.


starbelly:
Note: I also removed OTP 20 and elixir 1.14 from the test matrix. It was failing, but also they are both terribly old now.

paulo-ferraz-oliveira:
> Note: I also removed OTP 20 and elixir 1.14 from the test matrix. It was failing, but also they are both terribly old now.

If you wanna do this (I don't think there's a maintenance range in place) I'd suggest getting rid of pre-20 info. in the README, too, as well as some of the unit tests (`setup-beam.test.js`).

action.yml
```diff
@@ -1,9 +1,7 @@
 ---
 name: setup-beam
 description: >
-  Set up a BEAM-based CI environment, with Erlang/OTP as a base followed, optionally, by other
-  elements, like Elixir, Gleam, `rebar3` or `hex`, on top of that. Check the [README](https://github.com/erlef/setup-beam)
-  for updated information.
+  Set up an Erlang/OTP based CI environment with support for languages and tooling such as Elixir, Gleam, rebar3, and `hex`.
```
paulo-ferraz-oliveira, line 4: I'd specifically (recently) changed to have "BEAM-based", because you told me searching for BEAM yielded no results in the Marketplace. 😕

paulo-ferraz-oliveira, line 4:
```suggestion
  Set up an Erlang/OTP based CI environment with support for languages and tooling such as Elixir, Gleam, Rebar3, and `hex`.
```

paulo-ferraz-oliveira: Since you don't like the `-based version 😄

starbelly: Yeah, I changed it to this because I was trying to cram all the things in, do you have a better suggestion?

paulo-ferraz-oliveira:
```
Set up a BEAM-based CI environment, supporting languages and tools such as Erlang/OTP, Elixir, Gleam, `rebar3`, and `hex`.
```

😄

README.md
```diff
@@ -189,7 +189,7 @@ jobs:
     name: Erlang/OTP ${{matrix.otp}} / rebar3 ${{matrix.rebar3}}
     strategy:
       matrix:
-        otp: ['20.3', '21.3', '22.2']
+        otp: ['21.3', '21.3', '22.2']
```
paulo-ferraz-oliveira, line 192: We could probably remove this. It's a repeat.

README.md
```diff
@@ -165,7 +165,7 @@ jobs:
     name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
     strategy:
       matrix:
-        otp: ['20.3', '21.3', '22.2']
+        otp: ['21.3', '21.3', '22.2']
```
paulo-ferraz-oliveira, line 168: We could probably remove this. It's a repeat.

action.yml
starbelly: That'll work!

README.md
starbelly: derpy derp.

paulo-ferraz-oliveira (approved):

PR #298, opened 2024-07-13
Fix usage of version `latest` not matching internal elements

### LINKED ISSUE #297: Receiveing 404 when using "latest" - Installing Erlang/OTP 27.0 - built on amd64/ubuntu-22.04
#### The bug

Currently, we're receiving 404 from multiple mirrors when trying to install the latest OTP.

https://github.com/elixir-cldr/cldr/actions/runs/9829732424/job/27135101507
https://github.com/elixir-cldr/cldr/pull/230/files

```
Run erlef/setup-beam@v1
  with:
    otp-version: latest
    elixir-version: latest
    hexpm-mirrors: https://builds.hex.pm/
  https://cdn.jsdelivr.net/hex
  
    github-token: ***
    install-hex: true
    install-rebar: true
    version-type: loose
    disable_problem_matchers: false
  env:
    CLDR_PATH: /home/runner/work/cldr/cldr/integration-test-9829732424-69-1
```

```
Installing Erlang/OTP 27.0 - built on amd64/ubuntu-22.04
  Action install Erlang/OTP 27.0 failed for mirror https://builds.hex.pm,/ with Error: Unexpected HTTP response: 404
  Action install Erlang/OTP 27.0 failed for mirror https://cdn.jsdelivr.net/hex, with Error: Unexpected HTTP response: 404
```

#### Software versions

A list of software versions where the bug is apparent, as detailed as possible:

* `setup-beam`: 1.18

#### How to replicate

An ordered list of steps to replicate the bug: you could re-run the CLDR workflow our try to use latest.

#### Expected behaviour

Should install Erlang/OTP 27.0

#### Additional context

It would be helpful if the action produced better logs; I would love it if I could see the exact URL, headers, etc.

Schultzer:
# Description

~Add integration test.~ Fix latest

@paulo-ferraz-oliveira, there are some flaky tests or a bug in the version resolution, which is unrelated to my changes.

Closes #297 

- [X] I have performed a self-review of my changes

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -436,6 +436,8 @@ function getVersionFromSpec(spec0, versions0) {
       // If `version-type: strict` or version is RC, we obtain it directly
       version = versions0[spec]
     }
+  } else if (spec0 === 'latest') {
+    version = versions0[versions0.latest]
   } else if (rangeMax !== null) {
     // Otherwise, we compare alt. versions' semver ranges to this version, from highest to lowest
     const thatVersion = spec
@@ -504,7 +506,7 @@ function isRC(ver) {
 }
 
 function isKnownBranch(ver) {
-  return ['main', 'master', 'maint', 'latest'].includes(ver)
+  return ['main', 'master', 'maint'].includes(ver)
 }
 
 function githubARMRunnerArchs() {
--- a/dist/index.js
+++ b/dist/index.js
@@ -9536,6 +9536,8 @@ function getVersionFromSpec(spec0, versions0) {
       // If `version-type: strict` or version is RC, we obtain it directly
       version = versions0[spec]
     }
+  } else if (spec0 === 'latest') {
+    version = versions0[versions0.latest]
   } else if (rangeMax !== null) {
     // Otherwise, we compare alt. versions' semver ranges to this version, from highest to lowest
     const thatVersion = spec
@@ -9604,7 +9606,7 @@ function isRC(ver) {
 }
 
 function isKnownBranch(ver) {
-  return ['main', 'master', 'maint', 'latest'].includes(ver)
+  return ['main', 'master', 'maint'].includes(ver)
 }
 
 function githubARMRunnerArchs() {


paulo-ferraz-oliveira:
👋, @Schultzer.

What tests in particular?

I know that e.g. the installation on Windows, for 25.2.5.2 fails (I detected it in another repo.) and it seems that we're trying to have 27.0 resolved instead of 27.0.1, which is something recent.

src/setup-beam.js
```diff
@@ -436,6 +436,8 @@ function getVersionFromSpec(spec0, versions0) {
       // If `version-type: strict` or version is RC, we obtain it directly
       version = versions0[spec]
     }
+  } else if (spec0 === 'latest') {
+    version = versions0[versions0.latest]
```
paulo-ferraz-oliveira, line 440: This, and the change below, do not seem related to the title of the pull request (talks about adding, not fixing). Why is this required?

.github/workflows/ubuntu.yml
```diff
@@ -22,6 +22,10 @@ jobs:
       fail-fast: false
       matrix:
         combo:
+          - otp-version: 'latest'
+            elixir-version: 'latest'
+            rebar3-version: 'latest'
+            os: 'ubuntu-latest'
```
paulo-ferraz-oliveira, line 28: This would mean that if e.g. OTP, Elixir or `rebar3` were to fail in `latest` the action's results would be affected (which we don't want). As an example, Erlang/OTP Windows 25.2.5.2 has issues installing (I don't know the origin of those), but that shouldn't affect the action's CI results (which I'm pretty sure the results for `windows.yml` will show). Having CI fail and discussing fixes upstream should not be the action's concern.

paulo-ferraz-oliveira:
@Schultzer, I don't think we'll accept the tests on `latest` based on what was exposed before: one dependency fails, the action's tests shouldn't be affected.

On the other hand, if your changes to the code are required we'll still consider them, given that the PR title is adjusted.

I'll briefly push to remove tests on Windows 25.2 and to fix the 27.0 comparison (since it was not being properly re`strict`ed).

paulo-ferraz-oliveira:
@Schultzer, opened https://github.com/erlef/setup-beam/pull/299, in the meantime, and will re-run the tests 3/4 times to make sure there's no obvious flakiness.

src/setup-beam.js
Schultzer: If you take a look at the issue. Then you will see that currently there is a bug causing it to return 404, since we didn’t have any integration test, this was not caught.


Before
Erlang/OTP 27.0

After
Erlang/OTP OTP-27.0.1


The changes ensures that the proper version name is used and are tested.

Please notice the missing OTP-.

.github/workflows/ubuntu.yml
Schultzer: What is your suggestion for an integration test?

Schultzer:
> ⚠️ @Schultzer, I don't think we'll accept the tests on `latest` based on what was exposed before: one dependency fails, the action's tests shouldn't be affected.
> 
> On the other hand, if your changes to the code are required we'll still consider them, given that the PR title is adjusted.
> 
> I'll briefly push to remove tests on Windows 26.2 and to fix the 27.0 comparison (since it was not being properly re`strict`ed).

Maybe it would be beneficial to move the latest integration test into it’s own action, which could run nightly, and provide automated status for when the latest are working or not.

Schultzer:
> 👋, @Schultzer.
> 
> What tests are flaky in particular? (**edit**: maybe it was the word "flaky" that got me thinking, since it should mean that upon a re-run there's a possibility the flaky test passes; this would happen with the Windows installer issue, for example - if Erlang/OTP was resolved and e.g. 26.2.5.3 was released for Windows -, but not the `27.0` assumption)
> 
> I know that e.g. the installation on Windows, for [26.2.5.2 fails](https://github.com/erlang/otp/issues/8666) (I detected it in another repo.) and it seems that we're trying to have 27.0 resolved instead of 27.0.1, which is something recent.

Well, I would consider the 27.0 test flaky, if in the future it could match on 27.0.1. Which looks like what has happen.


I have no idea why 26.2.5.2 is failing.

.github/workflows/ubuntu.yml
paulo-ferraz-oliveira: There is no need for integration tests at this level. You can use the mock `builds.txt` files you added, but otherwise I wouldn't rely on online elements.

paulo-ferraz-oliveira:
> Maybe it would be beneficial to move the latest integration test into it’s own action, which could run nightly, and provide automated status for when the latest are working or not.

Maybe, but that's not something we'll maintain. The purpose of that would eventually be to find issues with the latest versions' installers, or similar, not the action itself, which isn't the purpose of the action. An issue is opened next to `erlang/otp`, already, and that's most surely where the fix will happen: it's up to developers using `latest` to take that risk, not the action's.

paulo-ferraz-oliveira:
> Well, I would consider the 27.0 test flaky, if in the future it could match on 27.0.1.

Maybe it's semantics, but a flaky test is one that, upon running, will result in unpredictable results (either it passes now, or it passes later). That isn't the case with the test case I fixed, since it wouldn't ever pass, in the future, once `27.0.x` was released.

src/setup-beam.js
paulo-ferraz-oliveira: > since we didn’t have any integration test

It should be possible to write a test that doesn't rely on online elements (similar to what you proposed with the fixed `builds.txt` elements). I'd rather prefer that than rely on remote inconsistencies outside of the action's control.

.github/workflows/ubuntu.yml
paulo-ferraz-oliveira: ⚠️ this change has to be reversed, as we won't be doing "online" integration tests, at this moment. Preferably, but it's on you to decide about your own time and energy, we'd have them in the test suites, using the mock `builds.txt`.

.github/workflows/windows.yml
```diff
@@ -22,6 +22,10 @@ jobs:
       fail-fast: false
       matrix:
         combo:
+          - otp-version: 'latest'
+            elixir-version: 'latest'
+            rebar3-version: 'latest'
+            os: 'windows-latest'
```
paulo-ferraz-oliveira, line 28: ⚠️ this change has to be reversed, as we won't be doing "online" integration tests, at this moment. Preferably, but it's on you to decide about your own time and energy, we'd have them in the test suites, using the mock `builds.txt`.

src/setup-beam.js
paulo-ferraz-oliveira: There's this test:

```javascript
  spec = 'latest'
  expected = 'OTP-27.0'
  got = setupBeam.getVersionFromSpec(spec, matrix.otp['ubuntu-24.04'])
  assert.deepStrictEqual(got, expected)
```

Whatever it's not testing well, if there's missing elements, it should.

I don't see how a change to the code wouldn't change those tests results (that same function is being tested), but I didn't dig deep.

paulo-ferraz-oliveira (changes requested):
The following changes are requested:

* remove "integration tests" with online elements (i.e. reverse the changes made to `ubuntu.yml` and `windows.yml`)
* make sure that `getVersionFromSpec` is well tested for `latest` as the changed code should've provoked changes in the tests (or there's, otherwise, something missing I don't understand)

paulo-ferraz-oliveira:
With #299 merged, this pull request can be rebased and adapted to the requested code changes. There should be no tests with issues after the rebase.

.github/workflows/windows.yml
Schultzer: That has already been tested; what this was catching was getOTPVersions that might be broken. But as I said in my prior PR, I'm not looking to do any major refactoring here.

.github/workflows/ubuntu.yml
Schultzer: That has already been tested; what this was catching was getOTPVersions that might be broken. But as I said in my prior PR, I'm not looking to do any major refactoring here.

src/setup-beam.js
Schultzer: These tests have no issues. I understand that you do not prefer integration tests, but an integration test would have caught an issue with getOTPVersions. Decoupling code and uniting testing is good practice, but with integration tests, we would know if there was a hidden contract between the layers.

As I said in my prior PR, I do not want to do any significant refactoring, regardless of how beneficial it might be. I'm just trying to make incremental changes to improve this codebase's health.

Schultzer:
> The following changes are requested:
> 
> * remove "integration tests" with online elements (i.e. reverse the changes made to `ubuntu.yml` and `windows.yml`)
> * make sure that `getVersionFromSpec` is well tested for `latest` as the changed code should've provoked changes in the tests (or there's, otherwise, something missing I don't understand)

Suppose the integration test passes after the code changes. In that case, there is clearly not an issue in getVersionFromSpec but rather between the layers, which is what integration tests are helpful for. 

We can conclude that the issue is in getOTPVersions. Or rather the contract between them.

Only an integration test would have caught this. Regardless I have made the requested changes.

paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
Merged. Thanks.

paulo-ferraz-oliveira:
This is part of https://github.com/erlef/setup-beam/releases/tag/v1.18.1.

PR #299, opened 2024-07-13
Fix CI (don't use 26.2.5 on Windows; be strict when comparing versions)

paulo-ferraz-oliveira:
# Description

1. remove Windows 26.2.5 from tests: (resolved) 26.2.5.2 is [not installing properly](https://github.com/erlang/otp/issues/8666), but the action's CI shouldn't be affected for this.

2. properly simulate `version-type` `strict` when required to not have e.g. `27.0` identify as (recently released) `27.0.1` when that's unexpected.

## Motivation

The motivation for this change was the report in #298, which made me look into why we had issues with version resolution. (issue is not in code, but tests' expectations)

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -44,7 +44,7 @@ jobs:
           - otp-version: '24.0.2'
             rebar3-version: '3.16'
             os: 'windows-2019'
-          - otp-version: '26.2.5'
+          - otp-version: '26.1'
             rebar3-version: 'nightly'
             os: 'windows-2019'
           - otp-version: '23.0'
--- a/test/setup-beam.test.js
+++ b/test/setup-beam.test.js
@@ -183,15 +183,17 @@ async function testOTPVersions() {
     expected = 'maint-26'
     got = await setupBeam.getOTPVersion(spec, osVersion)
     assert.deepStrictEqual(got, expected)
-
     simulateInput('version-type', before)
+
+    before = simulateInput('version-type', 'strict')
     spec = '27.0'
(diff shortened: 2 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -38,7 +38,7 @@ jobs:
             elixir-version: '1.14.5'
             os: 'windows-latest'
             version-type: 'strict'
-          - otp-version: '25'
+          - otp-version: '25.1'
```
paulo-ferraz-oliveira, line 41: `25.2.5.2` (identified by `25` has an install issue). Maybe this should be reported next to Erlang/OTP. I searched their CI and don't know where the install is tested in GitHub CI (e.g. `windows-2022`).

.github/workflows/windows.yml
```diff
@@ -131,7 +131,7 @@ jobs:
       fail-fast: false
       matrix:
         combo:
-          - otp-version: '25'
+          - otp-version: '25.1'
```
paulo-ferraz-oliveira, line 134: Same as above.

test/setup-beam.test.js
paulo-ferraz-oliveira, line 1: All of these pertain to properly testing over `strict`, otherwise (as is the case) Erlang/OTP releases `27.0.1` and our tests break because we're expecting `27` to resolve to `27.0`.

Schultzer (approved):

starbelly (approved):

paulo-ferraz-oliveira:
This is part of https://github.com/erlef/setup-beam/releases/tag/v1.18.1.

PR #303, opened 2024-09-20
Remove "I have performed a self-review of my changes" from the pull request template

paulo-ferraz-oliveira:
# Description

We don't have a CLA but the guidelines should be help enough.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,5 +4,4 @@ A brief description of your changes.
 
 Closes #&lt;issue&gt;.
 
-- [ ] I have performed a self-review of my changes
 - [ ] I have read and understood the [contributing guidelines](/erlef/setup-beam/blob/main/CONTRIBUTING.md)


paulo-ferraz-oliveira:
Removing the "I have performed..." item from the template, as per our recent discussion, since the content seems to potentially be mis-understood.

.github/workflows/ubuntu.yml
```diff
@@ -55,7 +55,7 @@ jobs:
             os: 'ubuntu-22.04'
           - otp-version: '24'
             elixir-version: '1.16'
-            rebar3-version: '3'
+            rebar3-version: '3.23'
```
paulo-ferraz-oliveira, line 58: It automatically chose 3.24 so I'm pinning this down to have CI pass.

wojtekmach (approved):

paulo-ferraz-oliveira:
Thanks; merging...

paulo-ferraz-oliveira:
Released in v1, v1.18 and v1.18.2. Thanks.

PR #304, opened 2024-09-20
feat: setup custom retry with builtin node fetch

btkostner:
# Description

This action uses node 20, which now has the built in `fetch` implementation. We can use that instead of the `@action/http-client` package. This allows us to do our own retry implementation to fix the long standing issue #260.

Closes issue: #260

- [ ] I have performed a self-review of my changes

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -289,7 +289,7 @@ async function getOTPVersions(osVersion) {
       })
   } else if (process.platform === 'win32') {
     otpVersionsListings.forEach((otpVersionsListing) => {
-      jsonParseAsList(otpVersionsListing)
+      otpVersionsListing
         .map((x) => x.assets)
         .flat()
         .filter((x) => x.name.match(/^otp_win64_.*.exe$/))
@@ -346,7 +346,7 @@ async function getGleamVersions() {
   )
   const gleamVersionsListing = {}
   resultJSONs.forEach((resultJSON) => {
-    jsonParseAsList(resultJSON)
+    resultJSON
       .map((x) => x.tag_name)
       .forEach((ver) => {
         const gleamMatch = ver.match(/^v?([^ ]+)/)
@@ -365,7 +365,7 @@ async function getRebar3Versions() {
   )
   const rebar3VersionsListing = {}
   resultJSONs.forEach((resultJSON) => {
-    jsonParseAsList(resultJSON)
+    resultJSON
       .map((x) => x.tag_name)
       .forEach((ver) => {
         rebar3VersionsListing[ver] = ver
@@ -695,21 +695,6 @@ function parseVersionFile(versionFilePath0) {
   core.endGroup()
 
   return appVersions
-}
-
-function jsonParseAsList(maybeJson) {
-  try {
-    const obj = JSON.parse(maybeJson)
-    if (!Array.isArray(obj)) {
-      throw new Error('expected a list!')
-    }
-
-    return obj
-  } catch (exc) {
-    throw new Error(
-      `Got an exception when trying to parse non-JSON list ${maybeJson}: ${exc}`,
-    )
-  }
 }
 
 function debugLog(groupName, message) {
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -561,7 +561,10 @@ function getRunnerOSVersion() {
 
 async function getUrlResponse(url, headers, attempt = 1) {
   try {
-    const response = await fetch(url, { headers })
+    const response = await fetch(url, {
+      headers,
+      signal: AbortSignal.timeout(10000),
+    })
     const contentType = response.headers.get('content-type') || ''
 
     if (!response.ok) {
@@ -1057,6 +1060,7 @@ function debugLoggingEnabled() {
 }
 
 module.exports = {
+  get,
   getOTPVersion,
   getElixirVersion,
   getGleamVersion,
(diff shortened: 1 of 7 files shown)


paulo-ferraz-oliveira:
👋 Is there a mechanism (even if very dumb/simple) to try and test this?

btkostner:
@paulo-ferraz-oliveira Looks like GH can just pull directly from the branch and everything should work. So try instead of using `erlef/setup-beam`, do `btkostner/setup-beam@retry` and see if that works.

src/setup-beam.js
```diff
@@ -560,38 +559,50 @@ function getRunnerOSVersion() {
...
-    const response = await httpClient.get(url, headers)
-    if (response.statusCode >= 400 && response.statusCode <= 599) {
-      throw new Error(
-        `Got ${response.statusCode} from ${url}. Exiting with error`,
-      )
+    if (contentType.indexOf('application/json') !== -1) {
+      return response.json()
+    } else {
+      return response.text()
+    }
```
btkostner, line 575: It looks like the automatic JSON decoding was the only real feature we needed.

src/setup-beam.js
```diff
@@ -560,38 +559,50 @@ function getRunnerOSVersion() {
...
+async function getUrlResponse(url, headers, attempt = 1) {
+  try {
+    const response = await fetch(url, { headers })
+    const contentType = response.headers.get('content-type') || ''
 
-    if (pageIdx !== null) {
-      url.searchParams.append('page', pageIdx)
+    if (!response.ok) {
+      throw new Error(`Got ${response.statusCode} from ${url}`)
     }
```
btkostner, line 569: Just checking if the response is ok or not. We can get more detailed with status code checking if we need, but I don't see a reason.

src/setup-beam.js
```diff
@@ -560,38 +559,50 @@ function getRunnerOSVersion() {
...
-  async function getPage(pageIdx) {
-    const url = new URL(url0)
-    const headers = {}
-    const GithubToken = getInput('github-token', false)
-    if (GithubToken && url.host === 'api.github.com') {
-      headers.authorization = `Bearer ${GithubToken}`
-    }
+async function getUrlResponse(url, headers, attempt = 1) {
+  try {
+    const response = await fetch(url, { headers })
```
btkostner, line 564: Fetch will throw an error on timeout, so that's why we are wrapping. Open to other options / ideas to make this cleaner.

src/setup-beam.js
```diff
@@ -560,38 +559,50 @@ function getRunnerOSVersion() {
...
-      )
+    if (contentType.indexOf('application/json') !== -1) {
+      return response.json()
+    } else {
+      return response.text()
+    }
+  } catch (err) {
+    if (attempt <= MAX_HTTP_RETRIES) {
+      core.debug(`Error during fetch. Retrying in 1000ms: ${err}`)
+      await new Promise((resolve) => setTimeout(resolve, 1000))
```
btkostner, line 579: Extremely simple backoff mechanic. I'm not sure we need to get too overly complicated here.

paulo-ferraz-oliveira:
It'll work if CI passes, I'm not worried about that bit. I was thinking of what we could do to have reliable repeatable tests.

src/setup-beam.js
paulo-ferraz-oliveira: Try-catch should be Ok here. This is not a huge script, and the handling is isolated.

src/setup-beam.js
paulo-ferraz-oliveira: If we weren't getting more detail no more detail is required, at this moment. There's been requests to include more detail around errors, but that's for a different pull request.

src/setup-beam.js
paulo-ferraz-oliveira: I'd do exponential backoff with the timeout. You already have `attempt` there so it's a question of using some simple math.

btkostner: Added 👍

btkostner:
@paulo-ferraz-oliveira Added a test using the built in node server. Verified it is working as intended locally. I added a 10 second timeout to fetch calls, though the original code had a 3 minute socket timeout by default which seems _very_ long.

https://github.com/actions/toolkit/blob/6dd369c0e648ed58d0ead326cf2426906ea86401/packages/http-client/src/index.ts#L531

paulo-ferraz-oliveira (approved):
This looks good to me, in general. I'll let other repo collaborators review too, since I might've missed something. I'll add them as reviewers.

paulo-ferraz-oliveira:
D'you wanna be a pal and remove the [failing test](https://github.com/erlef/setup-beam/actions/runs/10967817024/job/30458171613?pr=304#step:3:46) from CI? It happens because we're targetting OTP 24 and `rebar3` 3.24 doesn't support it.

btkostner:
Looks like something is installing rebar3 and the action does not like that. I'd say it's another issue that should be addressed in another PR. Probably something about force overwriting files when installing.

wojtekmach:
Thank you!

I'll fix the test failure on main. We have this line for rebar:

```js
fs.mkdirSync(bindir)
```

and we're doing it twice on CI, once for `npm test` step and once for `.tool-versions test` step. I'll change it to:

```js
fs.mkdirSync(bindir, { recursive: true })
```

paulo-ferraz-oliveira:
Released in v1, v1.18 and v1.18.2. Thanks.

PR #316, opened 2025-01-15
Add ability to specify OTP architecture

### LINKED ISSUE #315: Support for installing 32-bit Elixir on Windows

I need to build a 32-bit Windows `mix release` of my project, but this Action only installs the 64bit version of Erlang/OTP and Elixir.

I would like the ability to specify which processor architecture of OTP/Erlang to install, when running on windows.

I am picturing something like:
```yaml
  blah:
    name: Create release
    runs-on: windows-2022
    steps:
    - name: Set up Elixir
      uses: erlef/setup-beam@v1
      with:
        version-type: strict
        otp-version: "27.2"
        elixir-version: "1.18.1-otp-27"
        architecture: "win32"
```

N/A

#### Additional context
I think this would require updating `install` for win32 opt:
https://github.com/erlef/setup-beam/blob/e5b66193a0ae8fd3e4c0296f0c7a1c7d36241444/src/setup-beam.js#L801-L802
to optionally use the `win32` version. I see that elixir releases appear to be platform agnostic, so no change is required there?

I see that for Linux bob builds of OTP are used, and bob only builds 64bit releases. So I don't think it's possible to easily support this option for Linux environments.

gilbertbw:
# Description

Closes #315 .

n.b. I have checked `npm test` and it is returning an exit code of 0, but there is a massive amount of debug output so I could not see any success/fail messages from testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,12 +73,12 @@ and Erlang/OTP.
 
 | Operating system | Erlang/OTP   | OTP Architecture | Status
 |-                 |-             |                  |-
-| `ubuntu-18.04`   | 17.0 - 25.3  | 32 bit           | ✅
-| `ubuntu-20.04`   | 21.0 - 27    | 32 bit           | ✅
-| `ubuntu-22.04`   | 24.2 - 27    | 32 bit           | ✅
-| `ubuntu-24.04`   | 24.3 - 27    | 32/64 bit        | ✅
-| `windows-2019`   | 21\* - 25    | 32/64 bit        | ✅
-| `windows-2022`   | 21\* - 27    | 32/64 bit        | ✅
+| `ubuntu-18.04`   | 17.0 - 25.3  | x86_64, arm64    | ✅
+| `ubuntu-20.04`   | 21.0 - 27    | x86_64, arm64    | ✅
+| `ubuntu-22.04`   | 24.2 - 27    | x86_64, arm64    | ✅
+| `ubuntu-24.04`   | 24.3 - 27    | x86_64, arm64    | ✅
+| `windows-2019`   | 21\* - 25    | x86_64, x86      | ✅
+| `windows-2022`   | 21\* - 27    | x86_64, x86      | ✅
 
 **Note** \*: prior to 23, Windows builds are only available for minor versions, e.g. 21.0, 21.3,
 22.0, etc.
--- a/README.md
+++ b/README.md
@@ -71,14 +71,14 @@ be the latest.
 This list presents the known working version combos between the target operating system
 and Erlang/OTP.
 
-| Operating system | Erlang/OTP  | Status
-|-                 |-            |-
-| `ubuntu-18.04`   | 17.0 - 25.3 | ✅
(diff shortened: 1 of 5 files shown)


starbelly:
This LGTM. Would like to see what @paulo-ferraz-oliveira .

README.md
```diff
@@ -71,14 +71,14 @@ be the latest.
...
-|-                 |-            |-
-| `ubuntu-18.04`   | 17.0 - 25.3 | ✅
-| `ubuntu-20.04`   | 21.0 - 27   | ✅
-| `ubuntu-22.04`   | 24.2 - 27   | ✅
-| `ubuntu-24.04`   | 24.3 - 27   | ✅
-| `windows-2019`   | 21* - 25    | ✅
-| `windows-2022`   | 21* - 27    | ✅
+| Operating system | Erlang/OTP   | OTP Architecture | Status
+|-                 |-             |                  |-
+| `ubuntu-18.04`   | 17.0 - 25.3  | 32 bit           | ✅
```
wojtekmach, line 76: I believe we are getting Ubuntu from https://github.com/hexpm/bob?tab=readme-ov-file#erlang-builds and we don’t have 32bit builds.

Instead of 32bit/64bit perhaps it is more useful to mention x86, x86_64 and arm64 given we have former two for Windows and latter two for Ubuntu (and eventually macOS)

gilbertbw: Sorry about the incorrect 32-bit arch for Ubuntu. I have done what you suggested.

gilbertbw:
@starbelly do you think the failing check is some transient issue with builds.hex.pm? Would you be able to rerun it?

starbelly:
> @starbelly do you think the failing check is some transient issue with builds.hex.pm? Would you be able to rerun it?

That indeed was the issue. This LGTM and @wojtekmach seemed to think so as well.

README.md
```diff
@@ -194,6 +194,24 @@ jobs:
...
+on: push
+
+jobs:
+  test:
+    runs-on: windows-latest
+    steps:
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: '26'
+          otp-architecture: '32'
```
wojtekmach, line 212: sorry I haven't responded sooner. Why is this `32` as opposed to `x86`/`x86_64`? I believe the latter would be more future proof as there is slowly arm64-windows support being added: https://github.com/erlang/otp/pull/8991/files.

That being said it is fine to wait for arm builds to show up and we could deprecate 32/64 in favor of x86/x86_64 (note: you're gonna be able to run x86_64/x86_64 on arm64 windows but not the other way around) but I don't see a downside in doing it right now.

starbelly: Hey @wojtekmach , no release has been cut, if we need to we can revert, but also sounds like it's a future proof issue.

paulo-ferraz-oliveira:
This change was released in [v1.19.0](https://github.com/erlef/setup-beam/releases/tag/v1.19.0), also tagged `v1.19` and `v1` (moved).

Issue #321, opened 2025-03-11
Add support for lates **stable** release


JanEricNitschke:
#### Is your feature request related to a problem?

I have a very small elixir project where i always wont to verify that it runs in the most recent Erlang/OTP release.
However it currently fails when using latests due to what i assume is some dependencies not yet support OTP-28.0-rc1

https://github.com/JanEricNitschke/TicTacToe/actions/runs/13796937078/job/38590707182#step:5:9

#### Describe the feature you'd like

So what i would love to have is the ability to specify that i want the latest version, but restricted to only full releases.

essen:
I am seeing the same in Cowboy where the `latest` resolves to `OTP-28.0-rc2` which is marked as a pre-release on GitHub, while the true latest is `OTP-27.3`.

https://github.com/ninenines/cowboy/actions/runs/14025947860/job/39279963839#step:3:12

starbelly:
Yeah, I would qualify this as a bug myself. Thanks for reporting. Will look at resolving soon, otherwise a PR is welcomed.

JanEricNitschke:
> Yeah, I would qualify this as a bug myself. Thanks for reporting. Will look at resolving soon, otherwise a PR is welcomed.

If you had a quick pointer where this could most likely be fixed then i would take a look

essen:
Somewhere around here I think: https://github.com/erlef/setup-beam/blob/main/src/setup-beam.js#L407-L412

paulo-ferraz-oliveira:
@starbelly, we could update the existing `latest` to not take `-rc` into account (I can't remember the details of the initial requirement for that, but it's in the Issues' section, for sure). We could have e.g. `latest` and `latest-rc`, because "rc" also seems like a good candidate for testing every now and then.

Schultzer:
As I wrote this feature initially, then I would consider not returning rc in latest a breaking change.

However, I still don’t understand what the issue is with breaking actions. As this was specifically developed with the intention of something breaking rather that be somewhere in cldr or any other place.
Even Ubuntu latest is breaking OTP. So there is precedent and IMO you should expect things might not work in the future. https://github.com/erlef/setup-beam/issues/318

I do see the benefit in getting the most recent major version. Today it is already possible to retrieve the latest minor and patch releases by only setting the major version. So there might be an opportunity to leverage that to set a minimum major version e.g +27

essen:
[...]

PR #323, opened 2025-03-24
Add support for latest release

JanEricNitschke:
# Description

I added an additional version specifier of `stable`, that works just like `latest` except that it ignores RC versions.

Another way to do this is to directly adjust `latest` to also ignore those, but the tests are currently very explicit about RCs being favoured over previous stable versions so i opted for this approach.

Closes [#&lt;issue&gt;](https://github.com/erlef/setup-beam/issues/321).

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -25932,7 +25932,7 @@ async function main() {
   await maybeInstallRebar3(rebar3Spec)
 
   // undefined is replaced by a function, post- main branch merge
-  const setupBeamVersion = 'afe5922'
+  const setupBeamVersion = '67834dc'
   core.setOutput('setup-beam-version', setupBeamVersion)
 }
 
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -52,7 +52,7 @@ async function main() {
   await maybeInstallRebar3(rebar3Spec)
 
   // undefined is replaced by a function, post- main branch merge
-  const setupBeamVersion = 'afe5922'
+  const setupBeamVersion = '67834dc'
   core.setOutput('setup-beam-version', setupBeamVersion)
 }
 


JanEricNitschke:
Couldnt get the tests running locally despite setting the GITHUB_TOKEN, not sure whats wrong there

```
> test
> node test/setup-beam.test.js

::error::Tried to map a target OS from env. variable 'ImageOS' (got undefined), but failed. If you're using a self-hosted runner, you should set 'env': 'ImageOS': ... to one of the following: ['ubuntu18', 'ubuntu20', 'ubuntu22', 'ubuntu24', 'win19', '
win22']
::debug::Error fetching from https://api.github.com/repos/erlang/otp/releases?per_page=100&page=2: Error: Unauthorized
::debug::Error during fetch. Retrying in 2000ms
::debug::Error fetching from https://api.github.com/repos/erlang/otp/releases?per_page=100&page=1: Error: Unauthorized
::debug::Error during fetch. Retrying in 2000ms

```

wojtekmach:
Id change the meaning of “latest” to “stable”, ie not introduce a new concept.

GitHub releases can be marked as latest, eg:

https://api.github.com/repos/erlang/otp/releases/latest

and I think it is useful to maintain parity with that.

this would remove ability to get latest published RC, as it works right now, but honestly I’m not sure if it is that useful. If people want to test against the bleeding edge they probably should test against main branch anyway.

curious what others think!

starbelly:
> Id change the meaning of “latest” to “stable”, ie not introduce a new concept.
> 
> 
> 
> GitHub releases can be marked as latest, eg:
> 
> 
> 
> https://api.github.com/repos/erlang/otp/releases/latest
> 
> 
> 
> and I think it is useful to maintain parity with that.
> 
> 
> 
> this would remove ability to get latest published RC, as it works right now, but honestly I’m not sure if it is that useful. If people want to test against the bleeding edge they probably should test against main branch anyway.
> 
> 
> 
> curious what others think!

I fully agree with this. Maybe latest-stable?

JanEricNitschke:
So just rename what is currently `latest` to `latest-stable` and have it have the functionality if what i introduced as `stable`?

Because neither versions exactly match the GH release that is marked as latest, as per the readme entry for the current `latest`.

christhekeele:
> this would remove ability to get latest published RC, as it works right now, but honestly I’m not sure if it is that useful. If people want to test against the bleeding edge they probably should test against main branch anyway.
> 
> curious what others think!

My 2 🪙, I find this behaviour useful, though not a common case. I have a library that uses private Elixir compiler internals, and I test against edge Elixir to get ahead of breaking changes. Right now I use master, but my preference would be to just run on RCs, to ignore false positives that a breakage is coming, so it is useful to me.

Only started looking into changing over to RCs today, but will hold off until a decision is made here. 🙂

starbelly:
> So just rename what is currently `latest` to `latest-stable` and have it have the functionality if what i introduced as `stable`?

Yeah, it's a bit tricky, there's ambiguity in both terms, which led to this to begin with. I suppose it comes down to expectations, which is also why I suggested perhaps latest (current behavior) and latest-stable or just stable. Just stable on its own is a bit more ambiguous, but we shouldn't aim for perfection either. 

Would like to hear what people think of that.

christhekeele:
While I like the terminology "stable", I'll point out that to me, "latest" means ***latest release***, which stands in contrast to a pre-release/release candidate. So both feel analogous in this context, and it would be confusing to use both. My vote would be for a "latest" vs "edge" dichotomy.

JanEricNitschke:
I think "stable" and "edge" make sense to me. The most important part should be that they are properly documented.

JanEricNitschke:
I have now made it so that `edge` works like the current `latest` and specified that as the main way that this should be used in the readme, although `latest` is still an (deprecated) alias for that, as noted in the readme.

starbelly:
I think I'm good on stable and edge. I'd like to hear what @wojtekmach thinks.

JanEricNitschke:
Fixed the main error from the CI

JanEricNitschke:
Not sure about the failure of `[ubuntu / OTP 27, Elixir , Gleam 1.5.1, rebar3](https://github.com/erlef/setup-beam/actions/runs/14282057719/job/40032934522?pr=323)` that seems to be unrelated to my changed.

The unit test failure changes i assume come from [this](https://github.com/erlef/setup-beam/pull/323/files#diff-86f99337fc28b990485e054ebd09d9d3c29ec10775791de7abd2f1a85e17e241R850) test case, but i dont quite know where exactly.

JanEricNitschke:
Ok, i fixed the first issue with the `object`, which was just that i had `stable` and `edge` missing in the `isValidVersion` regex. However now i am running into issues that `isRC` is not working correctly with things like `OTP-27.0-rc1` because it requires the actual version pattern to start at the beginning of the string.

Unfortunately i am not sure if i can just remove that requirement?

paulo-ferraz-oliveira:
I'd still like to have the ability to test "edge" (e.g. not deprecating it as the PR suggests). Whatever naming comes out of it (parity with other resources, e.g. GitHub, as @wojtekmach mentions, is Ok, but we can also build on top of it).

So, whether it's `latest`, `stable`, `latest-stable`, `edge`, my expectations would be:

1. if we're to break current behaviour, it should be signalled in a MIGRATION document (or the README itself) and bump the action to 2.x
2. if we're to keep current behaviour, but extend it, documentation should reflect the new names appropriately (as the PR hints at)
3. it'd be nice to have a stable channel on testing updates to upcoming versions (it's not the same testing an RC as it is testing `master`, since `-rc` comes from a different branch)

dist/index.js
```diff
@@ -9489,7 +9489,7 @@ function gt(left, right) {
 
 function validVersion(v) {
   return (
-    v.match(/main|master|nightly|latest/g) == null &&
+    v.match(/main|master|nightly|latest|stable|edge/g) == null &&
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Also, would it be nice to issue a warning if `latest` was used instead of `edge`? This way developers could start updating their CIs and approaching future expectations.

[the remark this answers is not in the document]
paulo-ferraz-oliveira: This is not the file we'll be changing, if we are changing anything, but I guess it'll be easy to find the reference to the comments in `setup-beam.js`.

README.md
```diff
@@ -56,16 +56,21 @@ For pre-release versions, such as `v1.11.0-rc.0`, use the full version
...
 it knows (**note**: it is not the same as [GitHub considers it](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository)
 and some repositories might propose).
 
 If in doubt do a test run and compare the obtained release with the one you were expecting to
 be the latest.
 
+#### "Edge" versions
+
+Similar to `stable`, except that release candidates are also considered.
+Currently, the same behaviour can be achieved with the deprecated `latest` alias.
```
paulo-ferraz-oliveira, line 72:
```suggestion
To target these, set a tool's version to `edge` (same behaviour as deprecated `latest`).
```

paulo-ferraz-oliveira: ("Currently" will eventually stop making sense)

dist/index.js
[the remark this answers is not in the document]
JanEricNitschke: I have this

> #### "Edge" versions

in the readme

README.md
JanEricNitschke: Definitely. If you have a preferred README text i will adjust it to that.

JanEricNitschke:
@paulo-ferraz-oliveira 

The current idea for the PR is to add a `stable` tag that works exactly like the current `latest` but ignores RCs.

Also an alias for the current `latest` is added as `edge` while the original `latest` is deprecated. It would be to be removed at some point while keeping just `edge` with the same functionality.

wojtekmach:
I don't have a strong opinion about but I'm not 100% sold on "edge" meaning "latest published released including RCs". I mean, it can mean anything and be documented as such, but my immediate association would be with nightly builds, the bleeding _edge_, way more unstable than an RC.

@paulo-ferraz-oliveira mentioned in the ticket:

> we could update the existing latest to not take -rc into account (I can't remember the details of the initial requirement for that, but it's in the Issues' section, for sure). We could have e.g. latest and latest-rc, because "rc" also seems like a good candidate for testing every now and then.

A big +1 for `latest` and `latest-rc` from me. I think it's immediately obvious what's what. It's not ideal to change the meaning of `latest` but I really think that's what people expect of that value and I'm not sure it's worth to go through deprecation period just to end up with the same value anyway.

JanEricNitschke:
I am completely impartial to the naming and how you want the readme entry to look. I only care that there is the functionality o get the latest non-RC version

README.md
paulo-ferraz-oliveira: The one I added above, for example. You can commit my suggestion or do it locally, and push, whatever's better for you.

dist/index.js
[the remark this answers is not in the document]
paulo-ferraz-oliveira: Yeah. I'll resolve this comment, since I already proposed a change to the README, later. Thanks.

paulo-ferraz-oliveira:
I agree with @wojtekmach. `latest` and `latest-rc` are more explicit, even if that means a seemingly breaking change. But we can deal with versioning later. We can consider it a fix, too, since we're now updating the README...

JanEricNitschke:
So change the current (as in for this PR) `latest`/`edge` to `latest-rc` and the current `stable` to `latest`. Anything else i should change?

paulo-ferraz-oliveira:
I'm good with those, for the time being. I'll let others pitch in, too, though.

wojtekmach:
+1 from me!

JanEricNitschke:
Done (hopefully i didnt mess anything up in the rename)

Schultzer:
As I wrote this feature initially, then I would consider not returning rc in latest a breaking change.

However, I still don’t understand what the issue is with breaking actions. As this was specifically developed with the intention of something breaking rather that be somewhere in cldr or any other place. 

I do see the benefit in getting the most recent major version. Today it is already possible to retrieve the latest minor and patch releases bu only setting the major version. So there might be an opportunity to leverage that to set a minimum major version e.g >27

paulo-ferraz-oliveira:
I think that, even if it can be considered breaking, this comment by @wojtekmach supports our decision to have `latest` not include `-rc`:

> > https://api.github.com/repos/erlang/otp/releases/latest
>
> and I think it is useful to maintain parity with that.

For Erlang/OTP `latest` is 27.x, not 28.x (as still `-rc`, atm).

Where we [discussed this initially](https://github.com/erlef/setup-beam/issues/266) I asked "And would we consider release candidates?", and your reply was "RC is welcome but unnecessary"; it seems to us that having it clearer as `latest-rc` make for less confusion.

Schultzer:
> I think that, even if it can be considered breaking, this comment by @wojtekmach supports our decision to have `latest` not include `-rc`:
> 
> > > https://api.github.com/repos/erlang/otp/releases/latest
> > 
> > 
> > and I think it is useful to maintain parity with that.
> 
> For Erlang/OTP `latest` is 27.x, not 28.x (as still `-rc`, atm).
> 
> Where we [discussed this initially](https://github.com/erlef/setup-beam/issues/266) I asked "And would we consider release candidates?", and your reply was "RC is welcome but unnecessary"; it seems to us that having it clearer as `latest-rc` make for less confusion.

Correct, but I think we’re forgetting that the context of that comment was about getting the feature landed, and I was willing to not include RC if for some reason that would be technically impossible or resistant towards its. So it was still intentional that it included RC in the end.
 However I don’t have an issue if the community wants to change it, but I do think that if breaking changes are introduced that it should follow semver.

paulo-ferraz-oliveira:
I'm personally Ok to release:

- `1.x` 
  - considering the initial implementation had a "bug" around the implemented intention: it seems like it, if we take the README's "Pre-release versions are opt-in" into account, or
- `2.x` 
  - forcing many a CI element to be updated for something that:
    - won't "break" - except for expectations - more than actually using the `latest` as-is
    - most users of the action aren't using - I have no hard stats for this, but I'm making an educated guess, that most people try to have some sort of deterministic approach in their CI.

Schultzer:
Personelly I would prefer a major bump, since there is no bug in latest and we won't break anyone who is expecting rc to be included, people that would like the new behavior can easily upgrade to 2. where we would have both latest and latest-rc.

From my perspective then this is the best of both worlds, and what most developer would expect. 

After alll it is right in the readme and part of the test cases:

> "Latest" versions
>
> Set a tool's version to latest to retrieve the latest version of a given tool. The latest version is (locally) calculated by the action based on the (retrieved) versions it knows (note: it is not the same as [GitHub considers it](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) and some repositories might propose).
>
>If in doubt do a test run and compare the obtained release with the one you were expecting to be the latest.

https://github.com/erlef/setup-beam/blob/4b97eeaf23ad23397976bb25e8345cb633a53ee4/test/setup-beam.test.js#L826L843

```
  versions = {
    '27.0.0-rc3': '27.0.0-rc3',
    '27.0.0-rc2': '27.0.0-rc2',
  }
  spec = 'latest'
  expected = '27.0.0-rc3'
  got = setupBeam.getVersionFromSpec(spec, versions)
  assert.deepStrictEqual(got, expected)

  versions = {
    '27.0.0-rc3': '27.0.0-rc3',
    '27.0.0-rc2': '27.0.0-rc2',
    '27.0.0': '27.0.0',
  }
  spec = 'latest'
  expected = '27.0.0'
  got = setupBeam.getVersionFromSpec(spec, versions)
  assert.deepStrictEqual(got, expected)
```

starbelly:
I want to float an alternative that may satisfy all interested parties. Why not keep latest as is and add an additional option for excluding certain types of releases? 

This would keep latest as is and avoid a breaking change (behavioral) and give us a way to solve the original issue. After all, all proposed solutions add a new additional option, yet as stated this avoids concerns with breaking changes. 

Thoughts?

README.md
```diff
@@ -56,29 +56,34 @@ For pre-release versions, such as `v1.11.0-rc.0`, use the full version
...
+Similar to `latest`, except that release candidates are also considered.
+To target these, set a tool's version to `latest-rc`.
+
 ### Compatibility between Operating System and Erlang/OTP
 
 This list presents the known working version combos between the target operating system
 and Erlang/OTP.
 
-| Operating system | Erlang/OTP   | OTP Architecture | Status
-|-                 |-             | -                |-
```
paulo-ferraz-oliveira, line 75: I think the only broken thing was the space after `|` below "OTP Architecture". I'd like to keep as is, instead of aligning the code visually since, e.g. a future change might imply changing multiple lines, which won't be good.

README.md
```diff
@@ -88,14 +93,14 @@ and Erlang/OTP.
 Self-hosted runners need to set env. variable `ImageOS` to one of the following, since the action
 uses that to download assets:
 
-| ImageOS    | Operating system
```
paulo-ferraz-oliveira, line 91: I don't think this was broken. Why was it updated?

paulo-ferraz-oliveira:
> add a new additional option

This is only partially correct. The options are still the same (which is why `action.yml` is unchanged): it's the value that changes from `latest` to something else (or the semantics, depending on how we want to move forward).

@starbelly, mind you we already have **a lot** of options, for an action that's supposed to do one thing only (install a CI environment for OTP + optional language):

* otp-version
* otp-architecture
* elixir-version
* gleam-version
* install-hex
* install-rebar
* rebar3-version
* version-type
* disable_problem_matchers
* version-file
* hexpm-mirrors

(I'm aware most of them have defaults)

What would the scope of the option be? Let's suppose such an option is called `include-rc` (same applies for `exclude-...`). Should the fetched versions (as per the normal listing) now include `rc` too? Or would it just apply to `latest`? Because if it's there by default it'd be `include-rc` `false` since the normal listing doesn't include release candidates, and this would go against what we're discussing, since you'd have to have `include-rc` `true` and `latest` as value (to keep same behaviour) thus also being a breaking change.

I think:

1. most people don't use `latest`, as is - I have no data to support this
2. the ones that do might not expect it to include `rc` (because none of the installers have `rc` as their latest) - I have no data to support this
3. making most consumers bump to 2.x, reading a changelog just to figure out they have to do nothing, seems like overkill

@wojtekmach had a good suggestion about including a What's new in the release notes and keep at 1.x.

Thus:

1. we either "break" it per the above reasons and keep at 1.x
2. we move with 2.x

starbelly:
> Thus:
> 
>     1. we either "break" it per the above reasons and keep at 1.x
> 
>     2. we move with 2.x

I hear what you're saying. I think we can all agree has nice api has as few options as possible and just does the right thing 90% of the time. I'd like to figure out how to reduce our number of options, yet that would require a 2.x, and a digression from the convo. 

What I'm seeing though is the addition of a new option `latest-rc`, so this is what I meant when I said all solutions proposed (and written) thus far add a new additional option 😄. So my point was, if we're going to add that, why not just add an exclusion option and keep `latest` as is? Admittedly, it would be some what odd because there only differing type of release other than "stable" is indeed RC, so you would only end up with one type of release to exclude really 😄 

I will yield to consensus, but I did want to try and offer up a simple solution that will keep everyone happy, based on all the commentary that I was indeed following.  

Ergo, I'm ok with going with a breaking behavioral change if we are loud about it.

Schultzer:
> > Thus:
> > ```
> > 1. we either "break" it per the above reasons and keep at 1.x
> > 
> > 2. we move with 2.x
> > ```
> 
> I hear what you're saying. I think we can all agree has nice api has as few options as possible and just does the right thing 90% of the time. I'd like to figure out how to reduce our number of options, yet that would require a 2.x, and a digression from the convo.
> 
> What I'm seeing though is the addition of a new option `latest-rc`, so this is what I meant when I said all solutions proposed (and written) thus far add a new additional option 😄. So my point was, if we're going to add that, why not just add an exclusion option and keep `latest` as is? Admittedly, it would be some what odd because there only differing type of release other than "stable" is indeed RC, so you would only end up with one type of release to exclude really 😄
> 
> I will yield to consensus, but I did want to try and offer up a simple solution that will keep everyone happy, based on all the commentary that I was indeed following.
> 
> Ergo, I'm ok with going with a breaking behavioral change if we are loud about it.

I think the option is properly the best stop gap, until there is a clearer idea about what to do with the options in general, it might be best option (pun intended) since it forces us to think real hard about improve the overall API and deal with them for a 2. 

Introducing a breaking change or bump just for latest-rc seams to equally bad in hindsight.

@starbelly thank you for suggesting this alternative that can satisfy all, until there is a good consensus about what is a good API for 2.x. I don’t think we should rush this, and a stop gap unblock everyone without breaking anything. It’s fairly easy to add the option for your workflow if you don’t want rc in certain workflows.

Hindsight is always 20/20. I wish I saw this as clear as you!

paulo-ferraz-oliveira:
Ok, so let's try to think about the option... what would that be?, and how would it apply to both `latest` (which is not an option, but an option's value) and `-version` in general...?

Do you mean having a value like `latest-no/without-rc`? I thought this was discussed and refused, but I might not have understood it...

Schultzer:
> Ok, so let's try to think about the option... what would that be?, and how would it apply to both `latest` (which is not an option, but an option's value) and `-version` in general...?
> 
> Do you mean having a value like `latest-no/without-rc`? I thought this was discussed and refused, but I might not have understood it...
> 
> **Edit**:
> 
> > It’s fairly easy to add the option for your workflow
> 
> I don't think it is easy. As I stated above, if e.g. your option is `exclude-rc` this applies to `latest` Ok (would have to be `false` by default), but what would happen for the non-`latest` version calculations with that default? Behaviour would be inconsistent.

Something just hit me, maybe all of this is completely unnecessary, as we’re using https://www.npmjs.com/package/semver I believe you could simply define your version as >0 or >27 and get the latest major version. Keep in mind that I haven’t tested this and I don’t see anything in our test suite. I think it is important to verify if we could solve this by adding a test and some more documentation for this functionality.

paulo-ferraz-oliveira:
I'd be Ok, as per your latest reply, if:

1. `>ver` would mean `rc` are accepted (we'd clarify the doc specifically for this)
2. we would clarify the doc. for the fact that `latest` includes `rc`
3. we would clarify the doc. by stating "no ranges" don't include `rc` (basically the same as 1. but **written** for clarify)

Schultzer:
> I'd be Ok, as per your latest reply, if:
> 
> 1. `>ver` would mean `rc` are accepted (we'd clarify the doc specifically for this)
> 2. we would clarify the doc. for the fact that `latest` includes `rc`
> 3. we would clarify the doc. by stating "no ranges" don't include `rc` (basically the same as 1. but **written** for clarify)

1. When I read semver.js then to allow for rc with > then you have to specify it explicitly e.g >1.5.0-rc. 
2. it’s been a while but I actually believe due to how semver.js works, was the original reason for the inclusion of latest. So one could test against upcoming releases, and in that sense latest is semantically correct.
3. Yeah, I think it would be great expand on docs explicitly and with examples of when to use latest vs version range.

Regardless of all of these I think having proper tests for this functionality is requirement, and we can reference them in the readme, for anyone who might be in doubt of how the resolution works.

PR #327, opened 2025-04-25
Detect ubuntu version from /etc/os-release if ImageOS = Linux

### LINKED ISSUE #279: Identify Ubuntu version automatically?

Not related to a problem, but it would be nice if we could avoid having to set the `ImageOS` env var and the action could detect it for us (at least some of the time).

For Ubuntu-based runners, it would be great if the action could just read from `/etc/os-release` things like `ID=ubuntu` and `VERSION_ID=24.04` for example, rather than users setting `ImageOS=ubuntu24` explicitly.

Setting `ImageOS` works, just a thought for a small improvement 😄 

#### Additional context

Hopefully it makes sense and would be helpful!

iStefo:
# Description

Adds logic to `getRunnerOSVersion` to identify the ubuntu version by reading /etc/os-release.

I have extracted the function into its own file to avoid complications when importing it for unit testing.

Using fs.readFileSync is not ideal, but I saw that the codebase also makes use of other sync fs functions already. Overall, there are a few refactoring opportunities in the codebase 🙃

To avoid bringing in any new dependency, I made use of node's own test runner which is somewhat new and not perfect, but available on LTS. Node's test runner will run all files ending in `.test.js`, which includes the current single test file.

As test fixtures, I grabbed the contents of `/etc/os-release` for three ubuntu and one alpine version and added them to the test directory.

Closes #279.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "licenses": "yarn licenses generate-disclaimer > 3RD_PARTY_LICENSES",
     "markdownlint": "markdownlint *.md ./github/**/*.md",
     "shellcheck": "shellcheck .github/workflows/*.sh",
-    "test": "node test/setup-beam.test.js",
+    "test": "node --test",
     "yamllint": "yamllint .github/workflows/**.yml && yamllint .*.yml && yamllint *.yml",
     "clean-dist": "rm -rf ./dist",
     "lint": "npm run markdownlint && npm run shellcheck && npm run yamllint && npm run jslint",
--- a/src/lib/getRunnerOSVersion.js
+++ b/src/lib/getRunnerOSVersion.js
@@ -1,0 +1,85 @@
+const fs = require('node:fs')
+const core = require('@actions/core')
+
+const IMAGE_OS_TO_RUNNER_OS = {
+  ubuntu18: 'ubuntu-18.04',
+  ubuntu20: 'ubuntu-20.04',
+  ubuntu22: 'ubuntu-22.04',
+  ubuntu24: 'ubuntu-24.04',
+  win19: 'windows-2019',
+  win22: 'windows-2022',
+}
+
+function getRunnerOSVersion(imageOS, readFile = fs.readFileSync) {
+  core.startGroup('Get runner OS version')
+
+  core.info(`ImageOS is set to '${imageOS}'`)
+
+  if (imageOS === 'Linux') {
+    core.info('Attempting to detect OS version from /etc/os-release.')
+
+    try {
+      const osRelease = parseOSRelease(readFile)
+      core.debug('Parsed /etc/os-release: ' + JSON.stringify(osRelease))
+
+      const shortVersion = osRelease.VERSION_ID?.split('.')[0]
+      imageOS = osRelease.ID + shortVersion
+
+      core.info(`Detected '${imageOS}'`)
+    } catch (error) {
+      core.error('Failed to parse /etc/os-release: ' + error.message)
+    }
+  }
+
+  const containerFromEnvImageOS = IMAGE_OS_TO_RUNNER_OS[imageOS]
+  if (!containerFromEnvImageOS) {
+    throw new Error(
+      "Tried to map a target OS from env. variable 'ImageOS' (got " +
+        `${imageOS}` +
+        "), but failed. If you're using a " +
+        "self-hosted runner, you should set 'env': 'ImageOS': ... to one of the following: " +
+        "['" +
+        `${Object.keys(IMAGE_OS_TO_RUNNER_OS).join("', '")}` +
+        "']",
+    )
+  }
+
+  core.info('Resulting runner OS: ' + containerFromEnvImageOS)
+
+  core.endGroup()
+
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -33,4 +33,4 @@
   "engines": {
     "node": ">=20"
   }
-}
+}
```
paulo-ferraz-oliveira, line 36: This change does not seem to pertain to this pull request.

src/lib/getRunnerOSVersion.js
paulo-ferraz-oliveira, line 1: I'd prefer to keep this in a single file, for now. If the time comes to split we'll do a separate pull request. (also, this way we have to compare/review across different files, which is not ideal)

package.json
```diff
@@ -9,7 +9,7 @@
     "licenses": "yarn licenses generate-disclaimer > 3RD_PARTY_LICENSES",
     "markdownlint": "markdownlint *.md ./github/**/*.md",
     "shellcheck": "shellcheck .github/workflows/*.sh",
-    "test": "node test/setup-beam.test.js",
+    "test": "node --test",
```
paulo-ferraz-oliveira, line 12: I think this'll go away as soon as we move the new `lib/` file back into `setup-beam.js`...

src/lib/getRunnerOSVersion.js
paulo-ferraz-oliveira: Feel free to follow up with a refactor update, if you'd like, but lets keep concerns separated.

src/setup-beam.js
```diff
@@ -27,7 +29,7 @@ async function main() {
     versions = parseVersionFile(versionFilePath)
   }
 
-  const osVersion = getRunnerOSVersion()
+  const osVersion = getRunnerOSVersion(process.env.ImageOS)
```
paulo-ferraz-oliveira, line 32: I'd prefer to keep the `process.` stuff inside the function.

test/getRunnerOSVersion.test.js
```diff
@@ -0,0 +1,97 @@
+const { describe, it, afterEach } = require('node:test')
+const assert = require('node:assert')
+const { readFileSync } = require('node:fs')
+const { join } = require('node:path')
+
+const { getRunnerOSVersion } = require('../src/lib/getRunnerOSVersion')
+
+describe('getRunnerOSVersion', () => {
+  it('throws if no ImageOS env exists', async () => {
```
paulo-ferraz-oliveira, line 9: It'd be best if the tests followed the format we have today, even if it's more rudimentary; the reason for this is that not all maintaining this might be well versed in JavaScript and also because for me this is part of a bigger refactor. If you want to, list what you think could be improved (I think you hinted at it in your description) and we can deal with it later.

test/fixtures/os-release/alpine3.txt
paulo-ferraz-oliveira, line 1: I'd prefer to have these fixtures in the main test file, for now, until we decide on how to split the code base / improve it.

src/lib/getRunnerOSVersion.js
```diff
@@ -0,0 +1,85 @@
...
+  win19: 'windows-2019',
+  win22: 'windows-2022',
+}
+
+function getRunnerOSVersion(imageOS, readFile = fs.readFileSync) {
+  core.startGroup('Get runner OS version')
+
+  core.info(`ImageOS is set to '${imageOS}'`)
+
+  if (imageOS === 'Linux') {
```
paulo-ferraz-oliveira, line 18: Would it be better to just look at `/etc/os-release` existing than this? Reason for it is we seem to be moving from one constraint to the other. If we want to be more generic we can base ourselves on the standard (`/etc/os-release`) and not the results we're looking at today.

paulo-ferraz-oliveira:
I'm approving the workflow runs. Overall it's good, but I left comments where I think we can change less. For such a small change, 8 new files and syntax changes, etc. seems a bit much, but let's discuss.

src/setup-beam.js
paulo-ferraz-oliveira, line 1: I don't think you ran `npm run build-dist` (as https://github.com/erlef/setup-beam/blob/main/CONTRIBUTING.md proposes) so this'll likely fail CI.

test/getRunnerOSVersion.test.js
paulo-ferraz-oliveira: In the meantime, before discussion continues, I'm gonna see about moving the current tests to something more similar to this.

package.json
paulo-ferraz-oliveira: Check https://github.com/erlef/setup-beam/pull/328#issuecomment-2832603425.

test/getRunnerOSVersion.test.js
paulo-ferraz-oliveira: Check https://github.com/erlef/setup-beam/pull/328#issuecomment-2832603425.

paulo-ferraz-oliveira:
@iStefo, I don't mind taking this over or re-implementing it. I think it's a "good to have" functionality that fits nicely into the action. Lemme know...

PR #328, opened 2025-04-26
chore: approach Node.js' "suites and tests" way of testing

paulo-ferraz-oliveira:
# Description

This change stemmed from a proposal in https://github.com/erlef/setup-beam/pull/327.

The goal of the change is to approach a more modern (and readable?) way of writing tests and looking at execution output.

I didn't bother to review the existing tests, but only to propose a new format for writing them. If needed I can revise them in a different pull request, which means "This pull request is probably best seen with `Hide whitespace` enabled".

Ref.: https://nodejs.org/api/test.html#describe-and-it-aliases.

# Further considerations

* changes regarding Gleam aren't "this pull request" -specific, but they make "this pull request"'s CI fail, which is why I added them here. If you want I can do it in a separate pull request, though.

* changes regarding Ubuntu-24.04 aren't "this pull request" -specific, but they make "this pull request"'s CI fail (or get warned on), which is why I added them here. If you want I can do it in a separate pull request, though.


paulo-ferraz-oliveira:
@iStefo, feel free to give this a look-see and propose changes.

package.json
```diff
@@ -9,7 +9,7 @@
     "licenses": "yarn licenses generate-disclaimer > 3RD_PARTY_LICENSES",
     "markdownlint": "markdownlint *.md ./github/**/*.md",
     "shellcheck": "shellcheck .github/workflows/*.sh",
-    "test": "node test/setup-beam.test.js",
+    "test": "node --test --test-reporter=spec",
```
paulo-ferraz-oliveira, line 12: Without `--test-reporter=spec` `::debug::` gets output as `# ::debug::` which means GitHub Actions can't understand debug is off.

starbelly (approved):
LGTM ❤️

test/setup-beam.test.js
```diff
@@ -174,856 +159,847 @@ async function testOTPVersions() {
     'https://repo.hex.pm, https://cdn.jsdelivr.net/hex',
     { multiline: true },
   )
+  const previousRunnerArch = process.env.RUNNER_ARCH
```
iStefo, line 162: Configuring or restoring environment and mocks should general happen in `before`/`after` or `beforeEach`/`afterEach` blocks. This helps readers understand that setup/teardown code is being run (in contrast to code that dynamically generates test cases) and ensure that the modifications actually happen at the correct time (afaik the test runner doesn't necessarily execute the tests in the moment they are defined using `it)

iStefo (commented):
I appreciate you adopting the node test runner!

I think it greatly helps structuring tests, having a standardized way of configuring each test's preconditions (see my comment), identifying individual test failures and also allows executing specific parts of the suits in isolation (which is faster and avoids cluttering the the console with output from other tests).

So big 👍 from my side

test/setup-beam.test.js
paulo-ferraz-oliveira: Good to know, thanks. 👍 For now, I just wanted to have a base building block.

PR #330, opened 2025-05-22
Override OTP version when specified in Elixir spec

### LINKED ISSUE #329: Cannot use Elixir 1.18-otp-27 with OTP 28
#### The bug

Cannot use Elixir 1.18-otp-27 with OTP 28 as a combination

#### Software versions

A list of software versions where the bug is apparent, as detailed as possible:

* `setup-beam`: v1

#### How to replicate

An ordered list of steps to replicate the bug:

```yaml
      matrix:
        include:
          - elixir: '1.18.4-otp-27'
            otp: '28.0'

...

      uses: erlef/setup-beam@v1
      with:
        elixir-version: ${{ matrix.elixir }}
        otp-version: ${{ matrix.otp }}

```

```
Error: Requested Elixir / Erlang/OTP version (1.18.4-otp-27 / 28.0) not found in version list (did you check Compatibility between Elixir and Erlang/OTP?).Elixir and Erlang/OTP compatibility can be found at: https://hexdocs.pm/elixir/compatibility-and-deprecations.html
```

#### Expected behaviour

From https://hexdocs.pm/elixir/compatibility-and-deprecations.html

| Elixir version	| Supported Erlang/OTP versions |
| --- | ---|
| 1.18	| 25 - 27 (and Erlang/OTP 28 from v1.18.4)|

#### Additional context

- https://github.com/wemake-services/recase/pull/217

kevinschweikert:
# Description

When specifying an Elixir version like `1.18.4-otp-27`and OTP version `28.0` the action would overwrite this to `1.18.4-otp-28` which does not exist, although the versions are compatible. This fixes the issue by overriding the OTP major version when specified in the spec

Closes #329 .

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -184,7 +184,14 @@ async function getOTPVersion(otpSpec0, osVersion) {
 
 async function getElixirVersion(exSpec0, otpVersion0) {
   const otpVersion = otpVersion0.match(/^([^-]+-)?(.+)$/)[2]
-  const otpVersionMajor = otpVersion.match(/^([^.]+).*$/)[1]
+  let otpVersionMajor = otpVersion.match(/^([^.]+).*$/)[1]
+
+  const userSuppliedOtp = exSpec0.match(/-otp-(\d+)/)?.[1] ?? null
+
+  if (userSuppliedOtp && isVersion(userSuppliedOtp) && userSuppliedOtp !== otpVersionMajor) {
+    core.warning(`Elixir built for Erlang/OTP ${userSuppliedOtp} does not match the specified Erlang/OTP version ${otpVersionMajor}`)
+    otpVersionMajor = userSuppliedOtp
+  }
 
   const [otpVersionsForElixirMap, elixirVersions] = await getElixirVersions()
   const spec = exSpec0.replace(/-otp-.*$/, '')
--- a/test/setup-beam.test.js
+++ b/test/setup-beam.test.js
@@ -511,6 +511,12 @@ describe('.getOTPVersion(_) - Elixir', () => {
     got = await setupBeam.getElixirVersion(spec, otpVersion)
     assert.deepStrictEqual(got, expected)
 
+    spec = '1.16.2-otp-26'
+    otpVersion = 'OTP-27'
+    expected = 'v1.16.2-otp-26'
+    got = await setupBeam.getElixirVersion(spec, otpVersion)
+    assert.deepStrictEqual(got, expected)
+
     spec = '1.12.1'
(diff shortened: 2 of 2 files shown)


src/setup-beam.js
```diff
@@ -184,7 +184,14 @@ async function getOTPVersion(otpSpec0, osVersion) {
 
 async function getElixirVersion(exSpec0, otpVersion0) {
   const otpVersion = otpVersion0.match(/^([^-]+-)?(.+)$/)[2]
-  const otpVersionMajor = otpVersion.match(/^([^.]+).*$/)[1]
+  let otpVersionMajor = otpVersion.match(/^([^.]+).*$/)[1]
+
+  const userSuppliedOtp = exSpec0.match(/-otp-(\d+)/)?.[1] ?? null
+
+  if (userSuppliedOtp && isVersion(userSuppliedOtp) && userSuppliedOtp !== otpVersionMajor) {
+    core.warning(`Elixir built for Erlang/OTP ${userSuppliedOtp} does not match the specified Erlang/OTP version ${otpVersionMajor}`)
```
wojtekmach, line 192: I guess this is subjective but to me this is not a warning, it's a feature not a bug. Btw, does the warning appear in tests? do we have something like ExUnit.CaptureLog?

wojtekmach (approved):

src/setup-beam.js
kevinschweikert: Yes, the warning shows up in the test but also all debug messages. I found this https://github.com/actions/toolkit/issues/118 but it seems there is no built in way to capture the log for a single test case.

I have no preference if we want to handle it silently or warn the user that they have a mismatch. Just let me know what works best for you.

wojtekmach: Yeah, please remove the warning. We can always add it later.

paulo-ferraz-oliveira:
@wojtekmach, I'm Ok to ship it. I was just expecting CI to pass.

paulo-ferraz-oliveira (approved):

paulo-ferraz-oliveira:
What's probably failing here is that `npm run build-dist` (or similar) is not being executed - it's mentioned in the contributing guide, but maybe not explicitly so -, so the code used for testing would not be the one used for distributing.

paulo-ferraz-oliveira:
Argh, CI's now failing because of `rebar3` 3.25 (and nightly) it seems. I didn't dig into the details...

paulo-ferraz-oliveira:
I've created a draft (for now) PR: https://github.com/erlef/setup-beam/pull/331. This'll hopefully remove the issues with testing around the latest `rebar3`.

starbelly (approved):
LGTM

PR #331, opened 2025-05-25
Fix for `rebar3` `nightly`/`latest`

paulo-ferraz-oliveira:
# Description

Compatibility with Erlang/OTP versions lower than 26 is not guaranteed, so we try to move the versions forward / adapt them, when warranted to prevent issues around e.g. `rebar3` releases like the recent 3.25 one.


.github/workflows/ubuntu.yml
```diff
@@ -49,9 +49,9 @@ jobs:
             version-type: 'strict'
           - otp-version: '25'
             os: 'ubuntu-24.04'
-          - otp-version: '25'
+          - otp-version: '26'
             elixir-version: '1.16'
-            rebar3-version: '3'
+            rebar3-version: '3.25'
```
paulo-ferraz-oliveira, line 54: We pin it, so it doesn't catch us by surprise in the (near?) future.

.github/workflows/ubuntu.yml
```diff
@@ -63,12 +63,9 @@ jobs:
             rebar3-version: '3.18.0'
             os: 'ubuntu-latest'
             version-type: 'strict'
-          - otp-version: '25'
-            rebar3-version: 'nightly'
-            os: 'ubuntu-latest'
-          - otp-version: '26'
-            rebar3-version: 'nightly'
-            os: 'ubuntu-latest'
```
paulo-ferraz-oliveira, line 71: Since we're now going `latest`/`nightly`, this is not required.

.github/workflows/ubuntu.yml
```diff
@@ -145,11 +142,11 @@ jobs:
       fail-fast: false
       matrix:
         combo:
```
paulo-ferraz-oliveira, line 144: We benefit from `latest`, consistently, and thus this should break much less in the future.

.github/workflows/windows.yml
```diff
@@ -128,11 +125,11 @@ jobs:
       fail-fast: false
       matrix:
         combo:
```
paulo-ferraz-oliveira, line 127: We benefit from `latest`, consistently, and thus this should break much less in the future.

starbelly (approved):
LGTM

paulo-ferraz-oliveira:
This change was released in [v1.19.0](https://github.com/erlef/setup-beam/releases/tag/v1.19.0), also tagged `v1.19` and `v1` (moved).

wojtekmach:
👍

.github/workflows/windows.yml
```diff
@@ -38,15 +38,9 @@ jobs:
...
             version-type: 'strict'
-          - otp-version: '25'
-            rebar3-version: 'nightly'
-            os: 'windows-latest'
           - otp-version: '24.0.2'
             rebar3-version: '3.16'
             os: 'windows-2019'
-          - otp-version: '26.1'
-            rebar3-version: 'nightly'
-            os: 'windows-2019'
```
paulo-ferraz-oliveira, line 49: https://github.com/erlang/rebar3/pull/2954

PR #333, opened 2025-05-26
Allow passing (`mix`) version for `hex` and path for `rebar` installation

### LINKED ISSUE #275: Allow passing (`mix`) version for `hex` and `rebar3` installation

We are having timeouts fetching packages from hexpm, see here https://github.com/hexpm/hex/issues/1029

The possibility to pass the hex version to install with `mix local.hex --force` instead of implicitly using the latest.

paulo-ferraz-oliveira:
# Description

Allow specifying:

* `mix-hex-version` as a string (optional); used when `install-hex` is `true`
* `mix-rebar-path` as a string (optional); used when `install-rebar` is `true`; only available for local paths

Closes #275.


.eslintrc.yml
paulo-ferraz-oliveira, line 1: Replace by newer standard `eslint.config.mjs`

README.md
```diff
@@ -228,7 +229,7 @@ are found (i.e. `erl`, `erl.exe`, `rebar3`, `rebar3.exe`, ...).
 ### Elixir Problem Matchers
 
 The Elixir Problem Matchers in this repository are adapted from
-[here](https://github.com/fr1zle/vscode-elixir/blob/45eddb589acd7ac98e0c7305d1c2b24668ca709a/package.json#L70-L118).
+[vscode-elixir/package.json](https://github.com/fr1zle/vscode-elixir/blob/45eddb589acd7ac98e0c7305d1c2b24668ca709a/package.json#L70-L118).
```
paulo-ferraz-oliveira, line 232: this is `yamllint`

paulo-ferraz-oliveira:
I ended up doing quite a few things in the same PR. Lemme know if you prefer it broken up and I'll split it...

paulo-ferraz-oliveira:
I'm Ok with deprecating the previous option, in favor of `mix-hex-version` latest. If you notice it, this comes from an issue, so I just acted on the little context I had.

paulo-ferraz-oliveira:
I pushed an alternative; it's still fair to want to keep the previous behaviour (which the new alternative tries to complement by letting you specify a version), for which we'd close the linked issue (with an explanation) and we'd have rationale for future decisions in this area.

paulo-ferraz-oliveira:
In any case, don't close it, since if we don't move forward with the version changes there's other stuff I want to use from the PR, and I can revamp the title, description, etc...

paulo-ferraz-oliveira:
I'm gonna use this to create another pull request with everything except the options' change, which we're still deciding on...

PR #336, opened 2025-05-29
Minor maintenance updates

paulo-ferraz-oliveira:
# Description

I build on the experience from https://github.com/erlef/setup-beam/pull/333 to tweak some stuff. Commits tell the story via comments.

We:

- take care of some security issues issued by GitHub
- update our dependencies' versions to as latest as possible
- fix an issue with `yamllint` where the files weren't actually linted
- adapt to a more modern JS/`eslint` approach
- make `action.yml` more explicit with `required: true/false` where appropriate
- add `actionlint` to the list of action linters (and act on its results)
- act on lint results for more recent software versions

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -52,6 +52,16 @@ jobs:
       - name: Check if build left artifacts
         run: git diff --exit-code --ignore-space-at-eol
 
+  check_action:
+    name: Action
+    runs-on: ubuntu-latest
+    if: github.ref != 'refs/heads/main'
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Yeah, `dist` is always a problem, to me. On the one hand it's what the consumers actually use, on the other hand it's generated, so very difficult to review. I tend to rely on the linters, tests and analysers we have, mostly for security...

paulo-ferraz-oliveira:
This was released in [v1.20.0](https://github.com/erlef/setup-beam/releases/tag/v1.20.0).

PR #340, opened 2025-05-30
Dependabot version updates

### LINKED ISSUE #339: Would we benefit from auto-updates to dependencies?
I wouldn't mind having Dependabot update `erlef/setup-beam`'s dependencies.

1. we could use a scheduled approach
2. we could group dependencies (to reduce the number of PRs)
3. we wouldn't need to always be updating, since we won't always be releasing, but at least this would allow us to remain "fresh" (I sometimes update dep. versions by hand, but it's boring)

Stuff that'd benefit from these updates:

* `package.json` (and `package-lock.json` as follow-up)
* `.github/workflows`

paulo-ferraz-oliveira:
# Description

Allow for auto-updates on our dependencies' versions, via Dependabot.

Closes #339.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+---
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: /
+  schedule:
+    interval: weekly
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: weekly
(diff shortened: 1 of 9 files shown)


paulo-ferraz-oliveira:
This was released in [v1.20.0](https://github.com/erlef/setup-beam/releases/tag/v1.20.0).

paulo-ferraz-oliveira:
This was released in [v1.20.0](https://github.com/erlef/setup-beam/releases/tag/v1.20.0).

PR #341, opened 2025-05-30
Match only on what we know should match (versions start with numbers, after OTP-)

paulo-ferraz-oliveira:
# Description

Act on a weakness update suggested by CodeQL's scanning: https://github.com/erlef/setup-beam/security/code-scanning/23.

(at most 3 digits for major version)


paulo-ferraz-oliveira:
@starbelly, if you see this and agree with the content feel free to merge. Otherwise I'll act on comments later 😄 👋

PR #348, opened 2025-06-24
Add back deprecated runners with warning

paulo-ferraz-oliveira:
# Description

This broke for some consumers, but if we state (as a code warning, and by not having it in the `README.md`) that support is limited, this might not only nudge consumers in the right direction but also "support" those that want to work with deprecated elements.

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -26486,17 +26486,33 @@ function getRunnerOSVersion() {
     macos14: 'macOS-14',
     macos15: 'macOS-15',
   }
+  const deprecatedImageOSToContainer = {
+    ubuntu18: 'ubuntu-18.04',
+    ubuntu20: 'ubuntu-20.04',
+  }
   const containerFromEnvImageOS = ImageOSToContainer[process.env.ImageOS]
   if (!containerFromEnvImageOS) {
-    throw new Error(
-      "Tried to map a target OS from env. variable 'ImageOS' (got " +
-        `${process.env.ImageOS}` +
-        "), but failed. If you're using a " +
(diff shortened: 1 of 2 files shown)


petermm:
awesome!
just one blocker: `Error: core.warn is not a function` 

it's `core.warning` https://github.com/actions/toolkit/tree/main/packages/core#logging

have CI run(s) ready: https://github.com/petermm/AtomVM/actions/runs/15862621970/job/44723112273

but no hurry.

thanks!

paulo-ferraz-oliveira:
I've pushed a fix.

dist/index.js
```diff
@@ -26486,17 +26486,31 @@ function getRunnerOSVersion() {
...
-        "']",
-    )
+    const deprecatedContainerFromEnvImageOS =
+      deprecatedImageOSToContainer[process.env.ImageOS]
+    if (deprecatedContainerFromEnvImageOS) {
+      core.warning(
+        `You are using deprecated ImageOS ${deprecatedContainerFromEnvImageOS}. ` +
+          'Support for maintenance is very limited. Consider a non-deprecated version as ' +
+          'mentioned in the README.md.',
+      )
```
petermm, line 26502: we also need a `return deprecatedContainerFromEnvImageOS` inside the conditional, sorry for missing that.

paulo-ferraz-oliveira:
I've pushed a fix.

dist/index.js
paulo-ferraz-oliveira: `return` pushed.

petermm:
Thank you for accommodating this obscure usage of old OS/compilers!

PR #351, opened 2025-06-26
Fix calculating `-otp-` major for Elixir

paulo-ferraz-oliveira:
# Description

We keep the previous constraint of "user-defined version as 1.19-otp-28" winning over the OTP input, but we now use the `otpVersion` from the installed OTP instead of the user input, which also simplifies code around regular expressions.

## What changes

* we add `process.env.NODE_ENV` to allow bypassing the initial action run during tests
* instead of calculating `otpVersionMajor` based on input and regex, we make an actual system call (sync) to determine it and use it
  * the "bypass if forced" mechanism that was in place stays in place
* per @wojtekmach's input, the Elixir version matching mechanism was changes to start from the detected OTP versions and iterate down until it finds a corresponding Elixir version (this "avoids" OTP `master` / Elixir `main` from becoming Elixir `main` and not `main-otp-29`, for example, when that one's out
* I reduced duplicate (in nature) test for the Elixir version checks

... everything else remains the same

## Version matching examples (from the tests)

- OTP `<26` / Elixir `<1.17` results in Elixir `v1.16.3-otp-25`
- OTP `master` / Elixir `> 0` results in Elixir `v1.19-otp-28`
- OTP `master` / Elixir `main` results in Elixir `main-otp-28`
- OTP `OTP-27` / Elixir `1.18.x` results in Elixir `v1.18.4-otp-27`
- OTP `OTP-26` / Elixir `v1.15.0-rc.2` results in Elixir `v1.15.0-rc.2-otp-26`
- OTP `25.2` / Elixir `main` results in Elixir `main-otp-25`

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -10,9 +10,11 @@ const _ = require('lodash')
 
 const MAX_HTTP_RETRIES = 3
 
-main().catch((err) => {
-  core.setFailed(err.message)
-})
+if (process.env.NODE_ENV !== 'test') {
+  main().catch((err) => {
+    core.setFailed(err.message)
+  })
+}
 
 async function main() {
   checkOtpArchitecture()
@@ -1198,13 +1200,17 @@ function debugLoggingEnabled() {
 
 module.exports = {
   get,
-  getOTPVersion,
   getElixirVersion,
   getGleamVersion,
+  getOTPVersion,
   getRebar3Version,
   getVersionFromSpec,
   githubAMDRunnerArchs,
   githubARMRunnerArchs,
   install,
+  installOTP,
+  maybeInstallElixir,
+  maybeInstallGleam,
+  maybeInstallRebar3,
   parseVersionFile,
 }
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -30,15 +30,14 @@ async function main() {
     versions = parseVersionFile(versionFilePath)
   }
 
-  const osVersion = getRunnerOSVersion()
   const otpSpec = getInput('otp-version', true, 'erlang', versions)
   const elixirSpec = getInput('elixir-version', false, 'elixir', versions)
   const gleamSpec = getInput('gleam-version', false, 'gleam', versions)
   const rebar3Spec = getInput('rebar3-version', false, 'rebar', versions)
 
   if (otpSpec !== 'false') {
-    await installOTP(otpSpec, osVersion)
-    const elixirInstalled = await maybeInstallElixir(elixirSpec, otpSpec)
+    await installOTP(otpSpec)
+    const elixirInstalled = await maybeInstallElixir(elixirSpec)
     if (elixirInstalled === true) {
       const shouldMixRebar = getInput('install-rebar', false)
       await mix(shouldMixRebar, 'rebar')
@@ -58,7 +57,8 @@ async function main() {
   core.setOutput('setup-beam-version', setupBeamVersion)
 }
 
-async function installOTP(otpSpec, osVersion) {
+async function installOTP(otpSpec) {
+  const osVersion = getRunnerOSVersion()
   const otpVersion = await getOTPVersion(otpSpec, osVersion)
   core.startGroup(
     `Installing Erlang/OTP ${otpVersion} - built on ${getRunnerOSArchitecture()}/${osVersion}`,
@@ -80,10 +80,10 @@ async function installOTP(otpSpec, osVersion) {
   return otpVersion
 }
 
-async function maybeInstallElixir(elixirSpec, otpSpec) {
+async function maybeInstallElixir(elixirSpec) {
   let installed = false
   if (elixirSpec) {
-    const elixirVersion = await getElixirVersion(elixirSpec, otpSpec)
+    const elixirVersion = await getElixirVersion(elixirSpec)
     core.startGroup(`Installing Elixir ${elixirVersion}`)
     await doWithMirrors({
       hexMirrors: hexMirrorsInput(),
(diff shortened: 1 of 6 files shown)


src/setup-beam.js
```diff
@@ -192,13 +192,16 @@ function requestedVersionFor(tool, version, originListing, mirrors) {
   )
 }
 
-const knownBranches = ['main', 'master', 'maint']
+const knownBranches = ['master', 'maint', 'main']
```
paulo-ferraz-oliveira, line 195: Changed `main` to the end because it's part of `maint` and I wanted to do the replacement in a quick fashion, not overcomplicating the algo.

src/setup-beam.js
```diff
@@ -192,13 +192,16 @@ function requestedVersionFor(tool, version, originListing, mirrors) {
...
 const nonSpecificVersions = ['nightly', 'latest']
 
 async function getElixirVersion(exSpec0, otpVersion0) {
-  const otpVersion = otpVersion0.match(/^(?:OTP-)?(.+)$/)[1]
-  const regex = `^(\\d{1,3}|${knownBranches.join('|')}|${nonSpecificVersions.join('|')})?.*$`
-  let otpVersionMajor = otpVersion.match(new RegExp(regex))[1]
+  let otpVersion = otpVersion0.replace(/-rc.*/, '')
```
paulo-ferraz-oliveira, line 199: Step 1. remove `-rc...` references (suffix)

src/setup-beam.js
```diff
@@ -192,13 +192,16 @@ function requestedVersionFor(tool, version, originListing, mirrors) {
...
+  knownBranches.forEach((branch) => {
+    otpVersion = otpVersion.replace(`${branch}-`, '').replace(`${branch}`, '')
```
paulo-ferraz-oliveira, line 201: Step 2. remove `<branch>-...` references (prefix)

src/setup-beam.js
```diff
@@ -192,13 +192,16 @@ function requestedVersionFor(tool, version, originListing, mirrors) {
...
+  })
+  // after this we should have either a number-string of an empty string
+  let otpVersionMajor = otpVersion.match(/(\d{1,3}).*/)?.[1] ?? null
```
paulo-ferraz-oliveira, line 204: Step 3. fetch the major version as one of the first three digits.

src/setup-beam.js
```diff
@@ -192,13 +192,16 @@ function requestedVersionFor(tool, version, originListing, mirrors) {
...
 async function getElixirVersion(exSpec0, otpVersion0) {
-  const otpVersion = otpVersion0.match(/^(?:OTP-)?(.+)$/)[1]
-  const regex = `^(\\d{1,3}|${knownBranches.join('|')}|${nonSpecificVersions.join('|')})?.*$`
-  let otpVersionMajor = otpVersion.match(new RegExp(regex))[1]
+  let otpVersion = otpVersion0.replace(/-rc.*/, '')
+  knownBranches.forEach((branch) => {
+    otpVersion = otpVersion.replace(`${branch}-`, '').replace(`${branch}`, '')
+  })
+  // after this we should have either a number-string of an empty string
+  let otpVersionMajor = otpVersion.match(/(\d{1,3}).*/)?.[1] ?? ''
```
paulo-ferraz-oliveira, line 204: Step 3. fetch the major version as one of the first three digits or ''.

test/setup-beam.test.js
```diff
@@ -541,6 +541,26 @@ describe('.getOTPVersion(_) - Elixir', () => {
...
   it('returns the expected value', async () => {
+    const previousRunnerArch = process.env.RUNNER_ARCH
+    process.env.RUNNER_ARCH = 'X64'
+
+    spec = '<26'
+    // assuming this won't be bumped, since it's out-of-support from Erlang/OTP
+    expected = 'OTP-25.3.2.21'
+    if (process.platform === 'win32') {
+      // because this is online and Windows doesn't have the OTP- prefix
+      expected = '25.3.2.21'
```
paulo-ferraz-oliveira, line 552: For Windows, we get the results from the `releases` listing on GitHub so `OTP` is (for the time being) not there.

Lemme know if we need further tests in the `windows.yml`, though.

wojtekmach:
`otp-version: "> 0"` now picks the correct version. `elixir-version: "> 0"` still picks the RC.

I also noticed:

```
elixir-version: "> 0"
otp-version: master
```

doesn't work: https://github.com/wojtekmach/setup_beam_bug/actions/runs/15921976367/job/44910621256

unfortunately the action does not show the URL it's trying to download from.

This works though:

```
elixir-version: "> 0-otp-28"
otp-version: master
```

paulo-ferraz-oliveira:
I don't think I'll be able to write a proper test, not without some complexity, because `>` is a moving target, but I can try to fix it, still... for me, in the tests, it's picking `master` (OTP) and `v1.19` (Elixir).

paulo-ferraz-oliveira:
I think we'll have to open an exception for Elixir there:

`maint`, in Erlang, means https://erlangforums.com/t/clarify-the-role-of-the-maint-branch/4166/4?u=paulo-f-oliveira.

But I don't think there's a correspondence in Elixir's `builds.txt`, is there?

The same goes for `maint-...`.

For `master`, though, I can try to pin it to `master-otp-<latest>`, is that the expectation?

paulo-ferraz-oliveira:
> I was just trying to be helpful pointing error scenarios that were not caught with tests.

And they were helpful for sure, thank you.

> OTP maint-28 counterpart would be Elixir v1.18 branch and corresponding builds:

Is this easy to have a rule for? Or is it an unlikely scenario? Because I'm not sure we want exceptions like hardcoded versions...

> My understanding is OTP maintains compatibility for last 3 major versions only.

Correct, at least for the distribution protocol and support range.

---

What you're pointing out (as missing) is `master` / `maint` not pointing the correct version, right? Is there a rule we can apply here?

wojtekmach:
> > OTP maint-28 counterpart would be Elixir v1.18 branch and corresponding builds:

> Is this easy to have a rule for? Or is it an unlikely scenario? Because I'm not sure we want exceptions like hardcoded versions...

just to be clear, you were asking if there is a maint equivalent and there isn’t, and if there is maint-RELEASE and there is, vMAJOR.MINOR.

what do you mean by having a rule?

Are you asking what’s the rule `elixir-version: main, otp-version: master` should resolve the prebuilt elixir too? I think download otp master, get its version, 29, see if there is main-otp-29, there isn’t, try next in order, main-otp-28, bingo.

paulo-ferraz-oliveira:
> if there is maint-RELEASE equivalent and there is, vMAJOR.MINOR.

> what do you mean by having a rule?

I meant "if somebody were to choose OTP `maint-28` what Elixir version should the action choose"...

You wrote

> OTP maint-28 counterpart would be Elixir v1.18 branch and corresponding builds:

But then there's 4 versions, so which one should the version auto pick... I'm thinking e.g., for `> 0`. But if you think that in that case the choice should be done by the consumer then the action doesn't need to do anything different, I guess.

paulo-ferraz-oliveira:
> Are you asking what’s the rule elixir-version: main, otp-version: master should resolve the prebuilt elixir too? I think download otp master, get its version, 29, see if there is main-otp-29, there isn’t, try next in order, main-otp-28, bingo.

Per https://github.com/erlef/setup-beam/actions/runs/15926064004/job/44923493706?pr=351#step:5:8, I guess that's what'll happen when `otp-29` is out.

paulo-ferraz-oliveira:
I'm pushing a new test for OTP `master` / Elixir `main` to check the result...

wojtekmach:
> But then there's 4 versions

What are the 4 versions?

1.18.4
1.18.4-otp-27
1.18.4-otp-26
1.18.4-otp-25

those four?

wojtekmach:
FYI (sorry if obvious) precompiled elixir without otp release in the name is precompiled against the oldest OTP which that Elixir version supports.

On builds.hex.pm, 1.18.4 is the same file as 1.18.4-otp-25 and so is main and main-otp-25.

it would be nice if selecting elixir main and otp master would have picked main-otp-NEWEST (get all known main-otp-* and pick newest) but id say that’s a nice to have, it’s easy for users to set elixir-version: main-otp-28.

paulo-ferraz-oliveira:
I think I'm done for now, again. I'll re-summarise at the top.

paulo-ferraz-oliveira:
@wojtekmach, when you test this out (or just comment based on the content) I'll be ready to merge and release as patch.

PR #389, opened 2025-10-02
Compatibility matrix strategy

sebastiw:
# Description

Re-opened #313 with review fixes

Example on how one can maintain a compatibility matrix.
Note I've not tested keeping the os-version in the strategy itself, but according to the github documentation it works for the runs-on argument as well.

I'm currently running otp-rebar pinning for another project for OTP22-OTP27 versions below.
And I ran a similar thing (though not with erlef/setup-beam) previously for the same project for OTP17-21.

I'll add the os-version to my project for testing, but I wanted early feedback on your thoughts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,8 @@ and Erlang/OTP.
 
 ### Compatibility between Erlang/OTP and rebar3
 
-Check [version compatibility](https://github.com/erlang/rebar3?tab=readme-ov-file#compatibility-between-rebar3-and-erlangotp) in `erlang/rebar3`.
+Check [version compatibility](https://github.com/erlang/rebar3?tab=readme-ov-file#compatibility-between-rebar3-and-erlangotp)
+in `erlang/rebar3`.
 
 ### Self-hosted runners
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,10 @@ and Erlang/OTP.
 **Note** \*: prior to 23, Windows builds are only available for minor versions, e.g. 21.0, 21.3,
 22.0, etc.
 
+### Compatibility between Erlang/OTP and rebar3
+
+See [compatable versions](https://github.com/erlang/rebar3?tab=readme-ov-file#compatibility-between-rebar3-and-erlangotp) in rebar3 repo
```
paulo-ferraz-oliveira:
```suggestion
Check [version compatibility](https://github.com/erlang/rebar3?tab=readme-ov-file#compatibility-between-rebar3-and-erlangotp) in `erlang/rebar3`.
```

paulo-ferraz-oliveira:
If CI passes I'm merging this to the main branch, but not tagging yet (also, not really relevant, since this is a generic consideration, not part of the actual executed code).

paulo-ferraz-oliveira (changes requested):
Per CI output you should run `npm run build-dist`, fix until Ok, then push.

PR #457, opened 2026-04-21
Remove lodash Dependency

maennchen:
# Description

Removes lodash. NodeJS 24 contains RegExp.escape, no need for lodash.


paulo-ferraz-oliveira:
I like it. Don't know why tests fail, but no need to wait for reviews for such a small change, I'd say.

maennchen:
@paulo-ferraz-oliveira The tests seem to fail consistently. I don't really have the time to trace it down. Do you mind getting this over the line?

maennchen:
@paulo-ferraz-oliveira The issue was that there was a regex escape in a URL. This did not matter with lodash that apparently escaped a bit less, but mangled the URL using the built in escaping.

On failing tests:
* OTP latest fails because of a TLS issue
* And for Windows ARM, we missed a detail, fixed in #460

paulo-ferraz-oliveira:
I'll try to rebase this, when I'm done with the other pull requests, then move it forward, if I find time.

paulo-ferraz-oliveira:
Getting a bunch of these https://github.com/erlef/setup-beam/actions/runs/28327156197/job/83919004763?pr=457#step:3:16: "**Error**: Forbidden", but I don't yet know where they come from.

paulo-ferraz-oliveira:
They're coming from us consuming too much GitHub API, apparently. Will have to retry later.

paulo-ferraz-oliveira:
@maennchen the only failing checks are

image

I'm merging so I can tag a new version for the most recent changes. Thanks for the update (it seems I added `lodash` in the past, but I can't actually remember why...)

paulo-ferraz-oliveira:
Released in [`v1.24.1`](https://github.com/erlef/setup-beam/releases/tag/v1.24.1). The rolling tags `v1` and `v1.24` have been updated to point to this release.

PR #460, opened 2026-05-05
Add win11-arm64 to ImageOS mapping

maennchen:
# Description

Add win11-arm64 to ImageOS mapping

### DIFF
--- a/dist/index.js
+++ b/dist/index.js
@@ -55358,6 +55358,7 @@ function getRunnerOSVersion() {
   const ImageOSToContainer = {
     ubuntu22: 'ubuntu-22.04',
     ubuntu24: 'ubuntu-24.04',
+    'win11-arm64': 'windows-11-arm',
     win19: 'windows-2019',
     win22: 'windows-2022',
     win25: 'windows-2025',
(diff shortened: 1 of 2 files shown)


wojtekmach:
I believe we don’t have OTP builds for this target so I wouldn’t bother. I understand this PR is fixing tests but we shouldn’t be testing this in the first place in my opinion.

Windows arm64 can run x64 builds but they segfault. Or worse yet they do not but otherwise provide weird results. I believe this is caused by JIT because builds with it disabled have been ok in my tests.

maennchen:
Oh yes, you're right. Should I just disable that test in that case?

paulo-ferraz-oliveira:
I'm fixing some broken tests in https://github.com/erlef/setup-beam/pull/468. Per what @wojtekmach mentioned, and I validated in e.g. https://github.com/erlang/otp/releases we can instead remove the test. I'll do that and "document" the exception in the code.

paulo-ferraz-oliveira:
@maennchen, sorry for the late review. As mentioned here, and https://github.com/erlef/setup-beam/pull/468, it's probably best we remove the tests (as per the README) since we give no guarantees (or images) for that, by default.

Feel free to close, or continue discussing if you have further ideas around this.

paulo-ferraz-oliveira:

PR #461, opened 2026-05-07
fix: canonicalize Windows ImageOS variant suffixes

sh41:
# Description

Strip runner-image variant suffixes from Windows ImageOS values (e.g., "win25-vs2026" → "win25") to prevent label-to-image transitions from breaking the OS version lookup.

Ref: https://github.com/actions/runner-images/issues/14004

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,11 @@ jobs:
         ...
 ```
 
+For Windows images, suffix variants are canonicalized to the base `winNN` value.
+For example, `win25-vs2026` is treated as `win25`.
+This handles runner label-to-image issues discussed in
+[`actions/runner-images#14004`](https://github.com/actions/runner-images/issues/14004).
+
 ### Outputs
 
 The action provides the following outputs:
--- a/dist/index.js
+++ b/dist/index.js
@@ -55353,6 +55353,14 @@ function getRunnerOSArchitecture() {
   )
 }
 
+/**
+ * Strip Windows runner-image variant suffixes (e.g. "win25-vs2026" → "win25").
+ * Deliberately only targets Windows; see actions/runner-images#14004.
+ */
+function canonicalizeImageOS(imageOS) {
+  return imageOS.replace(/^(win\d{2})-.+$/, '$1')
+}
+
 function getRunnerOSVersion() {
   // List from https://github.com/actions/runner-images?tab=readme-ov-file#available-images
   const ImageOSToContainer = {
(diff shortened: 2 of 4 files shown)


dumbbell:
I'm also affected by this problem in https://github.com/rabbitmq/khepri/ and https://github.com/rabbitmq/horus/.

essen:
Please merge this to fix Windows support https://github.com/ninenines/cowboy/actions/runs/27927141608/job/82631663305

paulo-ferraz-oliveira:
I saw this now. Given another PR caught my attention first I started https://github.com/erlef/setup-beam/pull/468, to fix some tests, and eventually this was part of that. When that one gets merged, you can try to rebase to see if it makes sense and works in your use case. I went with explicit entries in the object, instead of "canonical" naming since I wasn't able to (limited time) find references that tell me that's actually a rule...

README.md
```diff
@@ -131,6 +131,11 @@ jobs:
         ...
 ```
 
+For Windows images, suffix variants are canonicalized to the base `winNN` value.
+For example, `win25-vs2026` is treated as `win25`.
+This handles runner label-to-image issues discussed in
+[`actions/runner-images#14004`](https://github.com/actions/runner-images/issues/14004).
+
```
paulo-ferraz-oliveira, line 138: Does this help the consumer of the action? Or is it "just" exposing internals?

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/pull/468 is merged.

Feel free to rebase and keep changes to:
- test/setup-beam.test.js

updating the rest as necessary. What I think we want to keep, for now, is the **explicit** list of OS, not a rule around them (for the naming part).

essen:
Could we have some default handling of unknown images so that unknown win = latest win, unknown mac = latest mac, and unknown ubuntu = latest ubuntu? Perhaps behind an option. That would help avoid breakage when GH changes things.

paulo-ferraz-oliveira:
> Could we have some default handling of unknown images so that unknown

Makes sense. On the other hand, wouldn't it be potentially hiding further issues?

How about we add a flag for the default, in that case, thus the "hiding" is the responsibility of the consumer?

essen:
Just defining one default wouldn't work for matrix jobs that cover multiple images, it has more chances to work if there's some small heuristic to detect Linux/Mac/Windows.

paulo-ferraz-oliveira:
Sure. We can try to add a visible warning, and act on `win`, `macos` and `ubuntu`. We can "fallback" to `...-latest` but mind you that sometimes this is `ARM64`, sometimes it's `x86_64`, by GitHub's own definition. e.g. Ubuntu latest is `x86_64`, while `macOS` is `ARM64`. You should create a new issue to track this, as it got deviated from the original intention.

paulo-ferraz-oliveira:

sh41:
Thanks for all the work on the action and feedback on this PR. The addition of an explicit lookup of `win25-vs2026` in #468 resolves the original issue for our use case. Combined with the preference for an explicit OS list rather than heuristics I think this PR is now redundant so I will close it.

Regarding the heuristics: I may not have phrased it very well, but the "canonicalization" was an attempt to collapse all of GitHub's randomly introduced variations on the `ImageOS` (`win25`, `win25-vs2022`, `win25-vs2026`, `win25-vsXXXX` etc) to a single "canonical" name recognized by `setup-beam` (`win25`) so that the next time GitHub do something similar, `setup-beam` doesn't break. See https://github.com/actions/runner-images/issues/14004 & https://github.com/actions/runner-images/issues/14017 for discussion and mentions from a variety of other code bases that broke because GitHub announced and implemented the change to the default windows-25 container's `ImageOS` value on the same day. 

Thanks again ❤️

paulo-ferraz-oliveira:
> I may not have phrased it very well, but the "canonicalization" was an attempt...

No, I got you, but maybe with a new option (or even default that warns on CI) it'll be easier to follow, in case it breaks.

PR #462, opened 2026-06-01
Add tests for arm64 Linux runner OS detection

jtdowney:
actions/runner-images#14113 gave arm64 Ubuntu images distinct ImageOS values (e.g. ubuntu24-arm64 instead of ubuntu24), which don't appear in the `ImageOSToContainer` table, causing setup to fail on ubuntu-*-arm runners. This has been fixed on main, but this PR adds tests to catch any regression.

### DIFF
--- a/src/setup-beam.js
+++ b/src/setup-beam.js
@@ -1306,6 +1306,7 @@ export default {
   getGleamVersion,
   getOTPVersion,
   getRebar3Version,
+  getRunnerOSVersion,
   getVersionFromSpec,
   githubAMDRunnerArchs,
   githubARMRunnerArchs,
--- a/dist/index.js
+++ b/dist/index.js
@@ -39137,6 +39137,7 @@ function debugLoggingEnabled() {
   getGleamVersion,
   getOTPVersion,
   getRebar3Version,
+  getRunnerOSVersion,
   getVersionFromSpec,
   githubAMDRunnerArchs,
   githubARMRunnerArchs,
--- a/test/setup-beam.test.js
+++ b/test/setup-beam.test.js
@@ -587,6 +587,44 @@ describe('OTP arch-specific install', () => {
   })
 })
 
+describe('.getRunnerOSVersion(_)', () => {
+  it('maps x86 ImageOS values to their container', async () => {
+    const previousImageOS = process.env.ImageOS
+
+    process.env.ImageOS = 'ubuntu22'
+    assert.strictEqual(setupBeam.getRunnerOSVersion(), 'ubuntu-22.04')
+
+    process.env.ImageOS = 'ubuntu24'
+    assert.strictEqual(setupBeam.getRunnerOSVersion(), 'ubuntu-24.04')
+
+    process.env.ImageOS = previousImageOS
+  })
+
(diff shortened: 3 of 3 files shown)


paulo-ferraz-oliveira:
Implementation seems Ok to me and no previous tests had to be changed, which is a good indicator of backward compatibility. I'll approve the tests for running, then you can check the results.

paulo-ferraz-oliveira:
Some tests are failing. I'll see which ones I can quickly fix for the main branch, after which you'll need to rebase.

src/setup-beam.js
```diff
@@ -676,10 +676,15 @@ function getRunnerOSVersion() {
     ubuntu18: 'ubuntu-18.04',
     ubuntu20: 'ubuntu-20.04',
   }
-  const containerFromEnvImageOS = ImageOSToContainer[process.env.ImageOS]
+  // actions/runner-images PR #14113 gave arm64 images distinct ImageOS values
+  // (e.g. ImageOS=ubuntu24-arm64 instead of ubuntu24). The CPU architecture is
+  // resolved separately via RUNNER_ARCH, so the arm64 variants map to the same
+  // OS container as their x86 counterparts.
+  const imageOS = (process.env.ImageOS || '').replace(/-arm64$/, '')
```
paulo-ferraz-oliveira: In trying to fix the tests I found myself having to do:

```diff
ubuntu22: 'ubuntu-22.04',
+ 'ubuntu22-arm64': 'ubuntu-22.04-arm64',
ubuntu24: 'ubuntu-24.04',
+ 'ubuntu24-arm64': 'ubuntu-24.04-arm64',
ubuntu26: 'ubuntu-26.04',
+ 'ubuntu26-arm64': 'ubuntu-26.04-arm64',
```

which is similar to your approach, albeit more explicit (less rules make it easier to add/maintain stuff, I believe, since I can't understand atm if GitHub will eventually break it or not...)

paulo-ferraz-oliveira: Not everything will get lost, since when you rebase, if you believe this is a better approach you can still propose it, and the added tests and comments are a plus.

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/pull/468 is merged.

Feel free to rebase and keep changes to:
- test/setup-beam.test.js

updating the rest as necessary. What I think we want to keep, for now, is the **explicit** list of OS (in `setup-beam.js`), not a rule around them (for the naming part).

Sorry for the conflicts, but we had some cleaning to do and those are a result of that (otherwise tests would still fail and prevent easing of future contributions).

paulo-ferraz-oliveira:

jtdowney:
I've updated the PR to just add the tests.

paulo-ferraz-oliveira:
Thank you for this one, too. I'll approve the workflow runs and merge if everything's ✅

paulo-ferraz-oliveira:
https://github.com/erlef/setup-beam/actions/runs/28439791679/job/84387813465?pr=462#step:5:15 seems to indicate you might've forgotten `npm run build-dist` on the latest commit.

jtdowney:
Indeed, I had, it is now fixed

paulo-ferraz-oliveira:
Thanks, once again.

